PR #63, opened 2026-04-06
Per-level LOD dimension dropping with CSR reduce

### LINKED ISSUE #62: LOD: per-level dimension dropping with CSR reduce LUT
## Problem

The LOD plan currently halves all LOD dimensions at every level and continues until every dimension reaches 1 element. This has two issues:

1. **Dimensions go below chunk size.** Once a dimension is smaller than its chunk, further halving is pointless — the chunk is partially filled and wastes space. The smallest useful extent is chunk_size.

2. **No control over aspect ratio.** For visualization, the pixel aspect ratio changing across LOD levels is fine (viewers are transform-aware). For other use cases it may not be. There's currently no way to choose.

### Stopping conditions

Three conditions, first one wins:

1. **Preserve aspect ratio** (optional, user-specified): stop when **any** LOD dimension reaches chunk size (≤1 chunk).
2. **Chunk size floor** (always): stop when **all** LOD dimensions reach chunk size.
3. **Max levels** (already exists): stop when `nlod >= max_nlod`.

Each dimension is clamped at chunk size: `next = max((cur + 1) / 2, chunk_size)`.

### Per-level LOD masks

When a dimension reaches chunk size, it drops from the LOD set for subsequent levels and becomes a fixed (batch) dimension. This means `lod_ndim`, `lod_to_dim`, and `fixed_dims_count` become per-level rather than global on the plan.

### Replace `ends` array with CSR reduce LUT

The current `ends` array stores contiguous ranges in child morton space for each parent element. This relies on the property that a parent's children form a contiguous block in child morton order — which only holds when all dimensions halve uniformly.

When a dimension drops from LOD to fixed, a z-slice of a 3D morton ordering is not contiguous in the child buffer. The `ends`-based approach breaks.

**Fix:** replace `ends` with a CSR-style LUT:
- `starts[i]` — offset into `indices` for destination element `i`  
- `indices[k]` — absolute position in the source buffer

The reduce inner loop becomes:
```c
for (j = starts[i]; j < starts[i+1]; ++j)
    accumulate(values[indices[j]]);
```

All morton math, batch mapping, and dimension-dropping complexity is absorbed into LUT construction (done once at init). The reduce at runtime is trivial.

**LUT sizes:** `starts` has `dst_lod_count + 1` entries per level. `indices` has `src_lod_count` entries per level (every source element appears exactly once). When a dimension drops out, `starts` grows by the dropped dimension's extent, but total `indices` count is unchanged. Similar memory footprint to today.

### What doesn't change

- **Gather** (linear → morton): per-level, uses that level's own LOD mask and shape. No cross-level concerns.
- **Morton-to-chunks** (morton → chunk pool): same — per-level operation.
- Only the **reduce** step bridges two levels and needs to handle the dimension transition.

### Performance

The reduce inner loop changes from sequential scan (`values[src_base + j]`) to indirect read (`values[indices[j]]`). With typical window sizes of 2–8 elements, this fits in L1 cache and should have negligible impact. The LUT construction cost is paid once at stream init, not per frame.

## Context

This was identified while building an acquire-zarr compatibility shim. acquire-zarr takes a different approach (stops dimensions independently, couples XY), which causes different problems (aspect ratio drift, refuses to build a pyramid for elongated images). Neither approach is ideal — this proposal addresses chucky's side.

nclack:
## Summary
- Dims clamp at chunk_size during halving (never go below)
- Per-level `lod_mask`/`lod_ndim`/`lod_to_dim`/`fixed_dims_count` — dims drop from LOD when they reach chunk_size
- CSR reduce LUT replaces contiguous `ends` array — handles dimension dropping correctly
- `preserve_aspect_ratio` config option stops pyramid when any dim reaches chunk_size
- Pre-computed CSR uploaded at init removes per-epoch `lod_fill_ends_gpu` call

Closes #62

## Test plan
- [x] GPU CSR reduce matches CPU reference for all dtypes and methods
- [x] `test_clamp_at_chunk_size` — no dim goes below chunk
- [x] `test_per_level_lod_masks` — verifies dimension drop transitions
- [x] `test_preserve_aspect_ratio` — early stop when any dim hits chunk
- [x] Benchmarks build and run properly


nclack:
still having some trouble with benchmarks on livescreen. file writing seems slow and maybe some bugs

PR #115, opened 2026-04-26
GPU writer_flush should block

nclack:
## Summary

Makes GPU `writer_flush()` block until all queued `pwrite_ref` jobs are durable, mirroring the CPU contract. Fixes a metadata-ordering bug where `update_append` was advertising chunks before they were on disk.

Background: PR #111 added `shard_sink_drain` to GPU destroy, but `stream_flush_body` itself was still a kick — it called `finalize_shards` and `update_append` then returned without waiting on the sink io_queue. CPU's `cpu_stream_flush_body` already waits on per-slot fences (`src/cpu/stream.body.c:309-312`), so this brings GPU to parity.

## Changes

- `src/writer.{h,c}`: New `shard_sink_drain_many` helper. Two-phase fan-out (record → wait) across N sinks; falls back to per-sink drain on calloc failure.
- `src/gpu/stream.c` (`stream_flush_body`): Drain after `finalize_shards`, before `update_append`. Mirrors CPU's drain → finalize → metadata ordering.
- `src/multiarray/stream.gpu.c` (`flush_impl`): Fan-out drain across all array sinks via the new helper.
- `src/multiarray/stream.gpu.c` (`multiarray_tile_stream_gpu_destroy`): Replaced inline two-phase block with the helper. Behavior preserved.
- `tests/test_flush_drain_gpu.c`: Regression test. Mirrors `test_destroy_drain.c` but spawns a flush thread and verifies it blocks on a sentinel io_queue job.
- `tests/test_flush_drain_multiarray_gpu.c`: Direct multi-sink test of `shard_sink_drain_many` with two `zarr_array` sinks. Catches indexing/iteration bugs in the fan-out (skipped sinks, wrong stride) via an asymmetric pool-seq priming trick.

## Test plan

- [x] revert the `shard_sink_drain` insert in `stream_flush_body` → `test_flush_drain_gpu` fails. Restore.
- [x] Existing `test_destroy_drain` still passes (multiarray destroy refactor preserves behavior).
- [x] Negative regressions for `_many`: skipping the loop tail, or using `&evs[0]` instead of `&evs[i * LOD_MAX_LEVELS]` in the wait loop, both make `test_flush_drain_multiarray_gpu` fail.

### DIFF
--- a/src/gpu/stream.c
+++ b/src/gpu/stream.c
@@ -253,6 +253,9 @@ stream_flush_body(struct stream_engine* e, struct stream_context* ctx)
     }
   }
 
+  if (shard_sink_drain(ctx->sink, ctx->levels.nlod))
+    return writer_error();
+
   // Final metadata update using pre-emit chunk counts.
   if (ctx->sink->update_append) {
     const uint8_t na = dim_info_n_append(&ctx->dims);
--- a/src/multiarray/stream.gpu.c
+++ b/src/multiarray/stream.gpu.c
@@ -571,6 +571,38 @@ flush_impl(struct multiarray_writer* self)
   }
 
   ms->active = -1;
+
+  // Fan-out drain across all sinks so flush is a commit point: when it
+  // returns, every queued pwrite_ref job is durable and metadata reflects
+  // only durable chunks. stream_flush_body already drained per-array, but
+  // re-drain here to give callers a single barrier across the multiarray.
+  {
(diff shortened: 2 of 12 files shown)


nclack:
While trying to write a higher-level multiarray-flush integration test, found a hang in the multiarray-GPU + FS-pool flush path. Addressed in #117

PR #119, opened 2026-04-26
bench: match aqz benchmark precisely

nclack:
## Summary
- `bench_stream_aqz_single` now pins `chunk_size` and `chunks_per_shard` directly so the runtime layout matches the comment ("1 x 16 x 16 chunks/shard -> 128 MiB shards") and mirrors `benchmarks/benchmark.py` in the parent acquire-zarr repo.
- Previously the bench passed `min_shard_bytes = 128 MiB` + `target_concurrent_shards = 4` to the planner, and the planner satisfied the concurrency target by growing shards to 2 GiB (4 shards total, 4096 chunks/shard) — not the 64 × 128 MiB layout the comment claimed.

## Before vs. after on this host (CPU, --codec none, 8 GiB total)

| layout | shards | chunks/shard | shard bytes | wall GiB/s |
|---|---|---|---|---|
| planner-chosen (old) | 4 | 4096 | 2 GiB | ~3.4 |
| pinned 1×16×16 (new) | 64 | 256 | 128 MiB | ~3.7 |

### DIFF
--- a/bench/bench_stream_aqz_single.c
+++ b/bench/bench_stream_aqz_single.c
@@ -2,13 +2,6 @@
 //   tyx = 1024 x 2048 x 2048 u16, 64-cube chunks, 1 x 16 x 16 chunks/shard
 //   -> 64 shards of 128 MiB each.
 // Used for thread-scaling and IO-impact studies against tensorstore.
-//
-// Layout is pinned: chunk_size and chunks_per_shard are set directly and
-// chunk_ratios is left NULL so bench_util's resolve_chunk_sizing skips the
-// planner (no dims_budget_chunk_bytes, no dims_set_shard_geometry). This
-// preserves the exact (1, 16, 16) shard layout that benchmark.py uses; the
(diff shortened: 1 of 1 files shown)


bench/bench_stream_aqz_single.c
```diff
@@ -1,6 +1,14 @@
...
+//   tyx = 1024 x 2048 x 2048 u16, 64-cube chunks, 1 x 16 x 16 chunks/shard
+//   -> 64 shards of 128 MiB each.
 // Used for thread-scaling and IO-impact studies against tensorstore.
+//
+// Layout is pinned: chunk_size and chunks_per_shard are set directly and
+// chunk_ratios is left NULL so bench_util's resolve_chunk_sizing skips the
+// planner (no dims_budget_chunk_bytes, no dims_set_shard_geometry). This
+// preserves the exact (1, 16, 16) shard layout that benchmark.py uses; the
+// planner's target_concurrent_shards/min_shard_bytes hints would otherwise
+// collapse it to 4 large shards.
```
nclack:
```suggestion

```

PR #138, opened 2026-05-14
Macro-aggregation scaffolding (slot_batches, cursors, slot_chunk_cap)

nclack:
Phase 2 infrastructure for stacking multiple compressed batches into one D2H slot. No behavior change — `batches_per_slot` stays at 1, cursors at 0. Phase 3 will activate by raising `batches_per_slot_cap` and advancing cursors per kick.

**Stacks on #137** — merge that first.

- `struct batch_slice_entry`: per-batch slice metadata (data/desc base + per-LOD segments)
- `aggregate_slot`: adds `slot_cursor`, `slot_desc_cursor`, `slot_batches[]`, `batches_per_slot_cap`
- `slot_chunk_cap = W / 128 + nlod` for compressed codecs; pass-through stays at `C_per_batch + nlod`
- `aggregate_batch_unified_async` takes `slot_data_base` / `slot_desc_base` — kernels operate on biased views (Phase 2 values are 0)
- Delivery loop is now `batches × LODs` (degenerate outer loop)
- Contiguous-mode D2H derives end position from `slot->slot_batches[last]` rather than `handoff->layout`
- TODO comments mark Phase 3 traps (per-batch dispatch, metrics aggregation, single io_event per slot)

PR #139, opened 2026-05-26
Phase 3 macro-aggregation: scaffolding + cap=16 for codec=none

nclack:
## Summary

Lets the GPU stack up to 16 batches into one output buffer before flushing them to host, instead of flushing after each one. The choice of which output buffer a batch lands in is made on-GPU per kick.

Start with `src/gpu/aggregate.cu::fit_decision_k` (chooses the target output buffer for each batch) and `src/gpu/stream.flush.c::post_kick_review` (host-side orchestration that issues a flush when a buffer closes).

**Where it activates.** Stacking is on for compressed codecs and for uncompressed-contiguous output. The one configuration that stays at one-batch-per-flush is uncompressed output written into shards with per-page tails — each batch there is sized to its worst-case envelope, so only one fits in a buffer anyway.

**Safety check before stacking.** When shards have per-page tails, the GPU carries leftover bytes from one batch into the next inside the same output buffer. If a shard would finish partway through a buffer, that carry would land in the wrong place and corrupt the shard. Each batch is checked against this before it's committed; if stacking would be unsafe, the buffer closes early and the batch starts a fresh one.

## Test plan

- `test_cross_validate_lod_dim0` — multi-LOD uncompressed output is byte-exact against the CPU reference, across batches with different active levels.
- `test_flush_orchestration::test_two_batch_cycle` — two batches actually share one buffer when the codec supports it.
- `test_d2h_zstd_double_buffer` — four-cycle compressed run exercising buffer reuse.

## Bench (256³, 256K, 32 frames, 8 runs, median wall)

| codec | one-per-flush | stacking on |
|--|--|--|
| zstd | 1.4005 s | 1.4291 s |
| none | 0.4802 s | 0.4693 s |

Stacking adds a small per-batch overhead today. Tuning it out is follow-up work; landing the intended design comes first.

Closes #138.


nclack:
abandoning this. also see #140

PR #144, opened 2026-06-12
Unify GPU engine init, teardown, and binding

nclack:
Step 2 of the migration in `docs/gpu-orchestration.md` ("Unify init/teardown
and multiarray binding"), following #143. Behavior-neutral; net −150 LOC
(+1021/−1171 across 21 files).

## Why

The single-array engine and the multiarray path built the same engine three
ways: multiarray hand-reimplemented stage allocation (including a
field-by-field near-copy of `compress_agg_destroy`), context-switched arrays
through a ~45-assignment bind/unbind checklist that taxed every engine
change, and `tile_stream_gpu_memory_estimate` re-stated every allocation by
hand — and had drifted.

## What

- **Group per-array engine state** — per-array state becomes
  `struct compress_agg_array` + `struct engine_array_state` {batch,
  pool_current, flush, lod, agg}. Bind/unbind drops from the ~45-assignment
  two-way checklist (~105 lines) to five whole-struct assignments plus two
  fixups (capacity upload, LUT-cache invalidate); multiarray's
  `bind_context` is now 7 lines. New per-array fields ride the swap
  automatically.
- **Unify engine init and teardown** — shared `stream_engine_init/destroy`,
  `engine_array_state_init/destroy`, and split
  `compress_agg_init_shared/array_init` used by both paths.
  `src/multiarray/stream.gpu.c`: 956 → 414 lines. The genuine single-vs-
  multiarray differences survive as five explicit parameters (sizing source
  via `engine_limits_accumulate`, tail-gate on/off, `sync_flush`, scatter
  metric label, per-array capacity upload at bind); everything else was
  accidental duplication. One real waste removed: multiarray allocated
  `d_compressed` even for CODEC_NONE (never read — aggregate reads the pool
  directly); both paths now skip it.
- **Derive gpu memory estimate from init** — the 190-line hand mirror is
  replaced by sums of the same sizing functions the allocator calls
  (`codec_device_bytes`, `aggregate_batch_slot_memory`,
  `compress_agg_memory_estimate`, `lod_state_device_bytes`,
  `shard_state_heap_bytes`, …), each placed next to its allocation.

## Estimate drift found (and fixed by derivation)

The old estimate modeled a previous generation of the allocator: it omitted
all five per-shard device tables, `d_tail_carry`, and the L0 morton-chunk
LUT; gated LUT sizing on K>1 (the buffers always exist); and hardcoded
`2*(rank−1)` where the actual lifted rank is `2*(rank−n_append)`.
Measured: `256cube_multiscale` device estimate 10.27 → 10.34 GiB and
`orca2_multiscale` 6.67 → 6.69 GiB (same LOD-block root cause); other
bench scenarios unchanged at display precision. Bench-chosen configurations
are identical before/after — reporting accuracy only.

## Evidence (L40, sm_89)

- `ctest -E "(s3)"`: 46/46 in RelWithDebInfo **and** Debug;
  `test_cross_validate` ×8 (determinism, zstd round-trip, page-aligned
  tail-carry all 8/8).
- Multiarray explicitly: `test_multiarray_gpu` both configs;
  `bench_stream_256cube_two_streams` zstd + none; Debug run clean of
  ordering asserts.
- Zero new build warnings.

## Commit guide

- `Group per-array engine state` — the one-struct bind.
- `Unify engine init and teardown` — shared construction; multiarray's
  duplicate init/destroy deleted.
- `Derive gpu memory estimate from init` — single source of truth for
  sizing; deletes the drifted mirror.

### DIFF
--- a/docs/gpu-orchestration.md
+++ b/docs/gpu-orchestration.md
@@ -1,6 +1,6 @@
 # GPU stream orchestration: target design and migration plan
 
-Status: accepted direction, migration in progress (step 1).
+Status: accepted direction, migration in progress (step 2).
 Base: main + #142. Related: #140, #141, PR #142, `gpu-output-slot-ledger.md`.
 
 ## Why
--- a/src/gpu/aggregate.cu
+++ b/src/gpu/aggregate.cu
@@ -188,6 +188,38 @@ gather_batch_k(const void* __restrict__ d_compressed,
 }
 
 // ---------------------------------------------------------------------------
+// Sizing (must mirror aggregate_layout_upload / aggregate_batch_slot_init
+// exactly — tile_stream_gpu_memory_estimate sums these)
+// ---------------------------------------------------------------------------
+
+extern "C" size_t
+aggregate_layout_device_bytes(const struct aggregate_layout* layout)
+{
+  if (layout->lifted_rank == 0)
+    return 0;
+  return layout->lifted_rank * (sizeof(uint64_t) + sizeof(int64_t));
+}
+
+extern "C" int
+aggregate_batch_slot_memory(uint64_t batch_covering_count,
+                            size_t comp_pool_bytes,
+                            size_t* device_bytes,
+                            size_t* host_bytes)
+{
+  const uint64_t C = batch_covering_count;
+  size_t temp = 0;
+  if (aggregate_cub_temp_bytes(C, &temp))
+    return 1;
+  *device_bytes = 2 * (C + 1) * sizeof(size_t) // d_permuted_sizes + d_offsets
+                  + comp_pool_bytes            // d_aggregated
+                  + temp;                      // d_temp
+  *host_bytes = comp_pool_bytes                // h_aggregated
+                + (C + 1) * sizeof(size_t)     // h_offsets
+                + C * sizeof(size_t);          // h_permuted_sizes
+  return 0;
+}
+
+// ---------------------------------------------------------------------------
 // aggregate_batch_slot_init
 // ---------------------------------------------------------------------------
 
--- a/src/gpu/aggregate.h
+++ b/src/gpu/aggregate.h
@@ -41,6 +41,15 @@ extern "C"
                                 uint64_t batch_covering_count,
                                 size_t comp_pool_bytes);
 
+  // Sizing mirrors of aggregate_layout_upload / aggregate_batch_slot_init,
+  // for the memory estimate.
+  size_t aggregate_layout_device_bytes(const struct aggregate_layout* layout);
+
+  int aggregate_batch_slot_memory(uint64_t batch_covering_count,
+                                  size_t comp_pool_bytes,
+                                  size_t* device_bytes,
+                                  size_t* host_bytes);
+
   // d_tail_bytes: persistent per-LOD device array of size_t[num_shards]; read
   //   by add_shard_bias_k for this batch's leading-tail accounting. The host
   //   uploads the post-delivery values after every batch.
--- a/src/gpu/compress.cu
+++ b/src/gpu/compress.cu
@@ -152,6 +152,23 @@ Fail:
   return 0;
 }
 
+// --- codec_device_bytes ---
+
+// Device bytes codec_init allocates for this configuration. Must mirror the
+// allocations below exactly — tile_stream_gpu_memory_estimate sums this.
+extern "C" size_t
+codec_device_bytes(enum compression_codec type,
+                   size_t chunk_bytes,
+                   size_t batch_size)
+{
+  size_t bytes = batch_size * sizeof(size_t); // d_comp_sizes
+  bytes += batch_size * sizeof(size_t);       // d_uncomp_sizes
+  if (type != CODEC_NONE)
+    bytes += 2 * batch_size * sizeof(void*); // d_ptrs
+  bytes += codec_temp_bytes(type, chunk_bytes, batch_size); // d_temp
+  return bytes;
+}
+
 // --- codec_init ---
 
 extern "C" int
--- a/src/gpu/compress.h
+++ b/src/gpu/compress.h
@@ -31,7 +31,8 @@ extern "C"
   // Query max compressed output size per chunk (no GPU allocation).
   size_t codec_max_output_size(enum compression_codec type, size_t chunk_bytes);
 
-  // Query nvcomp workspace temp bytes (no GPU allocation).
+  // Total device bytes codec_init allocates (size arrays, ptr table, nvcomp
+  // temp) — sizing mirror for the memory estimate (no GPU allocation).
   size_t codec_device_bytes(enum compression_codec type,
                             size_t chunk_bytes,
                             size_t batch_size);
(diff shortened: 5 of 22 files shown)


nclack:
Reviewed against upstream/main with a field-level audit of the bind swap, an allocation-parity pass over both init paths, and an independent L40 (sm_89) validation run. **Verdict: ship with nits.** Behavior-neutrality holds everywhere I could attack it.

## Bind-swap accounting

Enumerated every field the old `bind_context`/`unbind_context` checklist moved (upstream/main `src/multiarray/stream.gpu.c:147-250`) and placed each one:

- **Inside a swapped struct**: `batch.{epochs_per_batch,accumulated}` (per-array K preserved — `engine_array_state_init` seeds it from the array's own `cl`), `pools.current`, all of `flush_pipeline` (slots/masks/pending/pending_seq/pending_handoff/next_seq), the whole `lod_state` (the old selective copy-back of `append_accum.counts`/`element_capacity` is subsumed; `morton_offset` is init-set-only so whole-struct copy-back is equivalent), and all of `compress_agg_array` (nlod, per-LOD layouts, shard_state copy-back, total_shards, page_size, tail buffers, shards_begin/n_shards, h_shard_capacity).
- **Fixups**: `d_shard_capacity` re-upload, `lut_cache_valid = 0`, plus `d2h_deliver.shard_alignment` from ctx (third fixup; was old checklist line 162).
- **Deliberately shared and untouched**: pools.buf, staging, codec, `d_compressed`, agg[2] slots (+ `io_done` drained in unbind, unchanged), LUT buffers and cache arrays (invalidated, never stale-hit), `pool_epochs_scratch`/stride, the four shard device tables + three host shadows (rebuilt per kick), d2h events/drain_stream, lod_shared, ordering (gate never armed on multiarray — confirmed also true on main), metrics.
- **Now swaps but used to be shared**: only `flush_pipeline.current` — dead (never read anywhere; the fc index is `pools.current`), so a no-op. Old unbind's engine-side zeroing is covered by the `memset(&e->compress_agg.ar, ...)`.
- **Dropped**: descriptor `batch_active_count[]` (write-only on main) and `compress_agg_stage.max_batch_layout` (only ever read to derive the three maxima) — both verified dead.

No dropped field; nothing swaps that should persist.

## Allocation parity

Walked both paths' complete allocation sets old vs new: identical in size expression and placement, including the multiarray subtleties (codec at max chunk_bytes, `d_compressed = codec_batch_max * codec.max_output_size`, agg slots at `C_max = max_covering + max_nlod`, shared tables at `max_total_shards`, per-array tail buffers at actual `total_shards`, SYNC_MEMOPS attributes, memsets, gate condition). Confirmed the three intentional deltas and nothing else:

1. multiarray `d_compressed` skipped for CODEC_NONE — verified never read on main's multiarray path (`kick_compress_phase` aggregates from `pool_buf`);
2. the doubled `pool_epochs_scratch` alloc-then-realloc gone (same final size);
3. dead `mx.compressed_bytes`/`max_output_size` maxima dropped (computed-but-unused on main).

`engine_limits_accumulate` is strictly the union of the two old sizing computations, plus K>=1 / mul-overflow checks the multiarray path previously lacked (strictly safer).

## Five-parameter claim

Hunted for a sixth difference: timing-event seeding order, `d2h_deliver_init` taking 0 + bind-time alignment (init only stores it; bind precedes use), `e->pool_bytes` now set on the single path (read only by the multiarray switch zero), error-path ordering in create/destroy, gate-init timing relative to the engine sync, layout-upload phase ordering. All neutral. The five parameters are real and exhaustive as far as I can determine.

## Memory estimate

Each sizing mirror checked against its adjacent allocation: `codec_device_bytes`, `aggregate_batch_slot_memory`, `aggregate_layout_device_bytes`, `lod_state_device_bytes` (incl. L0 morton LUT and accumulators), `shard_state_heap_bytes` (incl. tail/footer pools), and `compress_agg_memory_estimate`'s shard tables + tail buffers — all match exactly. The drift list in the description checks out against main's old estimate code (per-LOD-slot model, K>1 LUT gate, `2*(rank-1)` vs lifted rank, missing tables/tail-carry).

## Independent validation (L40, sm_89, CUDA 13.1→13.0)

- `ctest -E "(s3)"`: **46/46 RelWithDebInfo** (58.7s) and **46/46 Debug** (92.2s).
- `test_cross_validate` ×8: all pass (`gpu_zstd_determinism`, `gpu_zstd_round_trip`, `gpu_page_aligned_tail_carry` 8/8 each).
- Multiarray hard: `test_multiarray_gpu` ×8 all pass; `test_multiarray_flush_mock` ×4 all pass; `bench_stream_256cube_two_streams` zstd and none (200 frames) clean; Debug runs show no ordering asserts — only the documented by-design dead-edge `pool_consumed` warnings on the sync path.
- Build warning sets byte-identical to main (15 pre-existing test/bench warnings; zero new).
- Estimate old-vs-new on four scenarios (separate main build, same node): auto-fit **chosen configs identical** everywhere. `256cube_multiscale` device 10.27 → 10.34 GiB (lod 2.81 → 2.87) as claimed; `256cube_single` (zstd and none) unchanged; `orca2_multiscale` moves 6.67 → 6.69 GiB (lod 1.29 → 1.30) — the description's "other bench scenarios unchanged at display precision" is slightly inexact, same root cause (LOD/lifted-rank drift), no config impact.

## Findings (ranked)

1. **Low — double-destroy on the create-failure path re-frees stale LOD device handles.** `engine_array_state_init` destroys internally on `Fail` (`src/gpu/stream.init.c:267`), then both owners run `engine_array_state_destroy` again (`tile_stream_gpu_destroy` → `&s->ar`; multiarray `destroy_array_descriptor` → `&desc->st`). `compress_agg_array_destroy` memsets and the mask frees null out, but `lod_state_destroy` (`src/gpu/stream.lod.c`) frees `d_full_shape`/`d_lod_shape`/`d_gather_lut`/`d_fixed_dims_offsets`/`append_accum.d_*` without zeroing them, so the second pass re-frees stale handles — CUWARN noise today, and a real re-free if an interleaved alloc ever reuses the address. One-line fix: `memset(&st->lod, 0, sizeof(st->lod))` after `lod_state_destroy` in `engine_array_state_destroy` (or null the pointers in `lod_state_destroy`), making it idempotent like its siblings.
2. **Nit — stale comment cross-references.** `src/gpu/stream.flush.c:135` and `src/gpu/flush.compress_agg.c:659` still say "(shard_tables in stream.engine.h)"; the tail-gate contract they cite now lives on `struct compress_agg_array`.
3. **Nit — orphaned doc comment in `src/gpu/compress.h:34`.** `codec_device_bytes` was inserted directly under the "Query nvcomp workspace temp bytes (no GPU allocation)" comment that belongs to `codec_temp_bytes`.
4. **Nit — `docs/gpu-orchestration.md:3`** still reads "migration in progress (step 1)"; this PR completes step 2.
5. **Observation** — `flush_pipeline.current` is dead on both sides of this PR (never read); fine to leave, but it now rides the swap and reads as load-bearing.

PR #146, opened 2026-06-12
Drain sink IO in CPU writer_flush

### LINKED ISSUE #145: CPU writer_flush returns with footer IO pending → use-after-free → on-disk corruption
Observed on PR #144 CI (cpu-tests, docker): one-off failure, passed on rerun
of the identical commit.

```
ERROR tests/test_cpu_zarr_fs.c:923:  shard 2 chunk 3: bytes don't match any expected tuple
ERROR tests/test_cpu_zarr_fs.c:929:  shard 2: matched 3 chunks, expected 4
FAIL test_unbuffered_intra_batch_gen (2 integrity errors)
```
(run https://github.com/acquire-project/chucky/actions/runs/27390423905, first attempt)

## Why this is filed as a bug, not test noise

- The failure mode is wrong bytes on disk for one chunk of the final shard —
  an integrity failure, not an assertion artifact.
- Not reproducible outside CI so far: 100/100 passes on both the PR head and
  main on an NFS-backed filesystem (same binary loop, one node). The CI
  environment differs in exactly the dimension this test exercises: O_DIRECT
  semantics on docker overlayfs vs a real filesystem.
- The test (added in #125) drives the unbuffered path's hardest sequence:
  mid-batch shard finalization, page-floor direct writes, sub-page tail
  bounce, async IO workers with write fences. A timing-dependent ordering
  hole here would present precisely this way — rare, environment-sensitive,
  one chunk's bytes stale or torn.
- Two ordering bugs of this class were recently confirmed and fixed on the
  GPU path (#140, #141). The CPU unbuffered tail/footer path has its own
  fence/reuse cycle (footer_buf reuse fenced by record_fence/wait_fence;
  tail bounce buffers) that has not had the same stream-ordered audit.

## Suggested next steps

- Audit the CPU unbuffered delivery path's buffer-reuse fences
  (write_footer's footer_buf reuse, tail bounce lifetime vs queued
  write_direct) the same way the GPU edges were audited.
- If the audit finds nothing, instrument the test to dump the mismatched
  chunk + expected tuples on failure and loop it in CI's docker image
  (overlayfs) rather than a native FS.

Unrelated to #144 (its only CPU-visible change is an uncalled sizing
function); filing so the signal isn't lost.

nclack:
Closes #145.

## What

`cpu_stream_flush_body` returned with the final-flush shard footers still
queued in the sink's IO queue. Those footer writes are `write_direct` jobs
that *reference* the stream-owned `footer_buf`; a subsequent
`tile_stream_cpu_destroy` freed that memory while the IO worker could still
be parked before the write — a use-after-free whose freed bytes land on
disk in the footer region (leading bytes = the last chunk's sub-page tail,
which is why #145's CI failure corrupted exactly one chunk while index/CRC
stayed intact).

Fix mirrors the GPU body (`shard_sink_drain` after the `finalize_shards`
loop — the GPU got this in #115/#111; the CPU body never did, and #125's
switch to buffer-referencing footers turned the missing drain from a
durability gap into corruption). This also makes CPU `writer_flush` a
commit point (parity with #115) and covers the multiarray CPU path, which
flushes per-array through the same body.

## Proof (full matrix on #145)

With a 20 ms worker delay at upstream/main: ASAN reports the
use-after-free deterministically (3/3); delay + `MALLOC_PERTURB_=85` →
**0/30 pass**, all failures in final-flush footers. With the fix: 100/100
under both amplifications, ASAN clean.

## Regression test

`tests/test_flush_drain_cpu.c` — CPU twin of `test_flush_drain_gpu.c`
(#115): injects a gated blocking job into the pool's io_queue, runs
`writer_flush` on a second thread, and asserts it does not return before
the gate releases (geometry leaves a partial shard so a footer is queued
behind the gate). Deterministically red on unfixed main, green with the
fix — no timing-luck dependence.

## Validation

- CPU-only build: full ctest 29/29 (incl. the new test); 200/200 loops of
  `test_cpu_zarr_fs`.
- Full GPU build (sm_89, L40): ctest 47/47; build warnings identical to
  main.

The error-path variant (flush failing between queueing a footer and the
drain, both backends) is tracked separately — see follow-up issue.

### DIFF
--- a/src/cpu/stream.body.c
+++ b/src/cpu/stream.body.c
@@ -322,9 +322,6 @@ cpu_stream_flush_body(struct cpu_stream_view* v)
       }
     }
 
-    // finalize_shards queues footer write_direct jobs that reference
-    // stream-owned footer_buf memory; stream destroy frees it, so flush
-    // must not return while those jobs are pending. Mirrors the GPU body.
     if (shard_sink_drain(v->sink))
       return writer_error();
 
--- a/src/cpu/stream.body.c
+++ b/src/cpu/stream.body.c
@@ -322,6 +322,12 @@ cpu_stream_flush_body(struct cpu_stream_view* v)
       }
     }
 
+    // finalize_shards queues footer write_direct jobs that reference
+    // stream-owned footer_buf memory; stream destroy frees it, so flush
+    // must not return while those jobs are pending. Mirrors the GPU body.
+    if (shard_sink_drain(v->sink))
+      return writer_error();
+
     float emit_ms = (float)(platform_toc(&emit_clk) * 1000.0);
     if (v->metrics)
       accumulate_metric_ms(&v->metrics->sink, emit_ms, 0, 0);
--- a/tests/CMakeLists.txt
+++ b/tests/CMakeLists.txt
@@ -176,6 +176,7 @@ target_compile_definitions(
 add_gpu_test_exe(test_zarr_fs_sink   test_zarr_fs_sink.c   LINKS CUDA::cuda_driver CUDA::cudart_static test_zarr_helpers stream Zstd::Zstd test_platform test_shard_verify test_voxel_encode)
 add_gpu_test_exe(test_destroy_drain  test_destroy_drain.c  LINKS CUDA::cuda_driver CUDA::cudart_static stream zarr_array store_fs store_api shard_pool_fs platform test_platform Zstd::Zstd chucky_log)
 add_gpu_test_exe(test_flush_drain_gpu test_flush_drain_gpu.c LINKS CUDA::cuda_driver CUDA::cudart_static stream zarr_array store_fs store_api shard_pool_fs platform test_platform Zstd::Zstd chucky_log)
+add_test_exe(test_flush_drain_cpu test_flush_drain_cpu.c stream_cpu zarr_array store_fs store_api shard_pool_fs platform writer test_platform chucky_log)
 add_gpu_test_exe(test_multiarray_flush_mock  test_multiarray_flush_mock.c  LINKS CUDA::cuda_driver CUDA::cudart_static multiarray_gpu stream writer chucky_log)
 add_test_exe(test_zarr_s3_sink   test_zarr_s3_sink.c   zarr_array ngff_multiscale zarr_group zarr_metadata store_s3 store_api lod_plan dimension platform_cmd chucky_log)
 add_test_exe(test_store_s3       test_store_s3.c       store_s3 store_api platform_cmd chucky_log)
--- a/tests/test_flush_drain_cpu.c
+++ b/tests/test_flush_drain_cpu.c
@@ -1,0 +1,199 @@
+// Regression test: writer_flush() on the CPU stream must drain sink IO
+// before returning. finalize_shards queues footer write_direct jobs that
+// reference stream-owned footer_buf memory; without the drain, stream
+// destroy frees that memory while the IO worker may still read it.
+
+#include "platform/platform.h"
+#include "store.h"
+#include "stream.cpu.h"
+#include "test_platform.h"
+#include "util/prelude.h"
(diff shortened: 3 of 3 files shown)


nclack:
**Verdict: ship.**

Reviewed against the GPU body and the sink contract; independently rebuilt and validated on a CPU node (RelWithDebInfo + ASAN), including a red-check with the fix reverted.

**Fix semantics**
- Placement is the exact GPU mirror (src/gpu/stream.c:274): finalize loop → drain → `update_append` → `sink->flush`. That order is right for crash consistency (shard IO settles before the final metadata exposes the new extents), and it restores the documented contract at src/writer.h:79 — sink `flush` is "called from writer_flush after outstanding shard IO has settled" — which the CPU body had been violating.
- `shard_sink_drain` (src/writer.c:117) is no-op-safe on sinks without fence ops, is a full drain of everything queued before the fence on the fs/S3 pools, and surfaces prior async write errors via the sticky `io_error` → `has_error`; the body checks the return.
- Coverage: `tile_stream_cpu_destroy`'s auto-flush runs this same body, so destroy-without-flush is covered too. Multiarray CPU flushes per-array through the body onto the shared pool queue — N drains, each needed before that array's footer bufs can be freed; same design the GPU multiarray already documents (src/multiarray/stream.gpu.c:755). Acceptable: cost is footer-sized IO at a once-per-stream commit point.
- Blocking: GPU flush has blocked here since #115, and no CPU caller relies on flush being quick or polls `pending_bytes` afterward (`peak_pending_bytes` is GPU-only).
- Error paths: when the drain itself fails, the fence wait still completes before `has_error` is read, so the queue is empty and destroy stays memory-safe; an error *before* the drain (e.g. `finalize_shards` failing with an earlier level's footer already queued) keeps the hole — agreed that's the follow-up, and this PR doesn't widen it.

**Test**
- Gate mechanism is sound: the injected job polls `io_queue_is_shutdown`, so pool destroy can't deadlock on an unreleased gate; `wait_for_done` bounds the post-release wait, so a regression fails rather than hangs. A false red is impossible with the fix in place (flush cannot return before the gate opens); a false green would need >200 ms scheduling starvation of ms-scale work — same window as the GPU twin. Faithful twin of test_flush_drain_gpu.c; registered via unconditional `add_test_exe` (and `src/cpu` is unconditional), so it's present in CPU-only and GPU builds.
- Optional twin-sync nit: the one path that leaves the gate unreleased is `test_thread_start` failure, where `zarr_array_destroy`'s teardown drain would block behind the gate; `atomic_store(&gate, 1)` at the top of Cleanup closes it (the GPU twin shares this).

**Validation** (cpu-turin node, conda toolchain)
- PR build: ctest 29/31 — both failures are the S3 tests (no MinIO on the node); `test_flush_drain_cpu` 20/20.
- Red-check (73b55b1 reverted, test kept): 0/20 — every run fails with "flush returned before sink IO drained".
- ASAN + 20 ms `pwrite_ref_fn` delay (the #145 amplifier): unfixed → 3/3 heap-use-after-free in `pwrite_ref_fn` on memory freed by `shard_state_destroy` (src/zarr/shard_delivery.c:67); fixed → 10/10 clean, no ASAN reports.

No-action observation: the drain sits inside the `emit_clk` window (src/cpu/stream.body.c:325-329), so the IO wait is attributed to `metrics->sink` on CPU — accurate, and more informative than the GPU body, which doesn't time it.

src/cpu/stream.body.c
```diff
@@ -322,6 +322,12 @@ cpu_stream_flush_body(struct cpu_stream_view* v)
       }
     }
 
+    // finalize_shards queues footer write_direct jobs that reference
+    // stream-owned footer_buf memory; stream destroy frees it, so flush
+    // must not return while those jobs are pending. Mirrors the GPU body.
```
nclack:
```suggestion
```

PR #148, opened 2026-06-12
Carry buffer ordering in resource pools

nclack:
Step 3 of the migration in `docs/gpu-orchestration.md` — **resource pools
with generation-carried ordering** — following #143 (edge table) and #144
(engine unification). This is the step the plan exists for: after it, a
stage cannot observe a pointer to a cycled buffer without the
synchronization that makes it safe.

## What

`src/gpu/pool.{h,c}` (226 LOC): a pool binds a cycled resource's per-slot
payload to the two declared edges that order its generations (`ready`:
producer→consumer; `consumed`: consumer→producer). `acquire` queues the
wait that makes the payload safe in that role before handing the pointer
out; `release` records that role's completion. Pointers are minted only by
the pool API (`gpu_pool_view`); `gpu_pool_at` exists solely for documented
host-ordered paths (sync flush, bind, teardown). Edge identity, instancing,
debug asserts, and stall metrics stay in `gpu_ordering` — pools draw edges
from the table and never own events. Host-published generations (#142's
tail gate) are the same API with a `GEN_COUNTER` ready edge; teardown is
`gpu_pool_release_all` (preserving #142's release-before-blocking-sync and
memops degradation exactly).

Four resource families migrated, one commit each, suite green at every
commit:

1. **Staging slots** — ingest's four cross-stream waits/records become
   produce/consume acquire/release; the append loop's previously-unmetered
   busy-poll is a host acquire.
2. **Chunk pool** — the #140 edge is now structural: `drain_kick_and_swap`
   must `gpu_pool_acquire_produce` before re-zeroing.
3. **Aggregate slots** — three facets (device slot, host mirror, chunk
   index) over the same payload, each guarding a different consumer;
   `flush_handoff` carries pool handles instead of a raw slot pointer.
4. **Tail state** — `arm_tail_gate` → consumer acquire (threshold advances
   even when disabled, so every kick drains exactly once); `finish_drain` →
   producer release, on failure exits too.

`GPU_EDGE_LOD_DONE` deliberately remains a plain edge (dual-use timing
event; a second producer into the chunk pool).

## Boundary proof

Direct edge operations outside `ordering.c`/`pool.c`: **three**, all the
deliberate LOD_DONE sites. No stage signature takes a raw device pointer
for a pooled resource (`compress_agg_input.pool_buf` is a `gpu_pool*`;
ingest/LOD take `gpu_pool_view`); payload field access outside init/destroy
/bind is gone.

## Mutation checks (the step's exit criterion)

Temporary, never-committed mutations on the final commit:

- Chunk-pool acquire dropped (re-zero without waiting POOL_CONSUMED — the
  #140 bug reintroduced): `gpu_zstd_determinism` fails **8/8 reps**.
- Tail generation wait dropped (gate never armed — the #141 bug
  reintroduced): `gpu_page_aligned_tail_carry` fails **8/8 reps**.

Both restored: green. The pools demonstrably carry the ordering, and the
regression tests stand guard.

## Validation (L40, sm_89)

- Full `ctest -E "(s3)"` green in RelWithDebInfo **and** Debug at each of
  the four family commits; zero new build warnings; Debug runs clean of
  ordering asserts.
- `test_cross_validate` 8/8 reps (determinism, zstd round-trip,
  page-aligned tail-carry); `test_multiarray_gpu` + two_streams bench both
  codecs.
- Perf, base vs PR, same node, interleaved 3 reps × 5 scenarios × 2 codecs:
  zstd within 0.3% everywhere; codec=none within run noise (worst first-rep
  outlier re-tested with warmup: −0.9%). No standing regression.

## Commit guide

- `Add resource pool API; migrate staging` — pool.{h,c} + first family.
- `Migrate chunk pool to pool API` — #140's edge becomes structural.
- `Migrate aggregate slots to pool API` — the three facets + handoff
  change.
- `Migrate tail state to pool API` — #142's gate behind the GEN_COUNTER
  kind.
- `Trim comments to why-only` — comment hygiene pass.

### DIFF
--- a/src/gpu/CMakeLists.txt
+++ b/src/gpu/CMakeLists.txt
@@ -13,6 +13,7 @@ add_src_lib(stream
         stream.lod.c stream.lod.h
         stream.internal.h
         ordering.c ordering.h
+        pool.c pool.h
         flush.compress_agg.c flush.compress_agg.h
         flush.d2h_deliver.c flush.d2h_deliver.h
         flush.handoff.h flush.helpers.h
--- a/src/gpu/flush.compress_agg.c
+++ b/src/gpu/flush.compress_agg.c
@@ -49,6 +49,16 @@ compress_agg_init_shared(struct compress_agg_stage* stage,
 {
   memset(stage, 0, sizeof(*stage));
   stage->ord = ord;
+  gpu_pool_init(
+    &stage->agg_pool, ord, GPU_EDGE_AGG_DONE, GPU_EDGE_SLOT_DRAINED);
+  gpu_pool_init(&stage->agg_host, ord, GPU_EDGE_D2H_DONE, GPU_EDGE_COUNT);
+  gpu_pool_init(
+    &stage->agg_index, ord, GPU_EDGE_CHUNK_INDEX_READY, GPU_EDGE_COUNT);
+  for (int fc = 0; fc < 2; ++fc) {
+    gpu_pool_bind(&stage->agg_pool, fc, &stage->agg[fc]);
+    gpu_pool_bind(&stage->agg_host, fc, &stage->agg[fc]);
+    gpu_pool_bind(&stage->agg_index, fc, &stage->agg[fc]);
+  }
 
   const uint32_t K = lim->epochs_per_batch;
   const uint64_t M = lim->codec_batch;
@@ -589,13 +599,14 @@ Error:
   return 1;
 }
 
-// GPU_EDGE_SLOT_DRAINED: aggregate must not overwrite agg[fc].d_aggregated
-// before the prior D2H on the same fc has read it (seeded at init).
+// The aggregate slot's produce-acquire (SLOT_DRAINED): aggregate must not
+// overwrite agg[fc] before the prior D2H on the same fc has read it.
 static int
 acquire_inputs(struct compress_agg_stage* stage,
                const struct compress_agg_input* in,
                CUstream compress_stream,
-               struct gpu_pool_view* pool_buf)
+               struct gpu_pool_view* pool_buf,
+               struct aggregate_slot** slot)
 {
   CHECK(Error,
         gpu_pool_acquire_consume(in->pool, in->fc, compress_stream, pool_buf) ==
@@ -605,9 +616,11 @@ acquire_inputs(struct compress_agg_stage* stage,
     CHECK(Error,
           gpu_edge_wait(
             stage->ord, GPU_EDGE_LOD_DONE, in->fc, compress_stream) == 0);
+  struct gpu_pool_view v;
   CHECK(Error,
-        gpu_edge_wait(
-          stage->ord, GPU_EDGE_SLOT_DRAINED, in->fc, compress_stream) == 0);
+        gpu_pool_acquire_produce(
+          &stage->agg_pool, in->fc, compress_stream, &v) == 0);
+  *slot = v.p;
   return 0;
 
 Error:
@@ -669,6 +682,7 @@ static int
 dispatch_aggregate(struct compress_agg_stage* stage,
                    const struct batch_aggregate_layout* layout,
                    int fc,
+                   struct aggregate_slot* slot,
                    CUdeviceptr d_aggregate_src,
                    CUstream compress_stream)
 {
@@ -684,7 +698,7 @@ dispatch_aggregate(struct compress_agg_stage* stage,
             layout->total_batch_covering,
             stage->ar.nlod,
             stage->codec.max_output_size,
-            &stage->agg[fc],
+            slot,
             stage->shards.d_base_offsets,
             stage->shards.d_shard_capacity,
             stage->shards.d_tps_group,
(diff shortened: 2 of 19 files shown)


nclack:
## Review: independent audit + revalidation on L40 (sm_89)

**Verdict: ship-with-nits.** Walked every acquire/release pair in all four
families against upstream/main, audited each `gpu_pool_at` site's
host-ordering claim, and independently re-ran the validation gates and both
mutation checks. No correctness holes found; findings below are comment
precision and one API-scope note.

### Pairing audit (per family)

- **Staging** (`stream.ingest.c`, `stream.c`): `dispatch_h2d` +
  scatter/multiscale consume resolve to the identical
  wait/record/stream/instance sequence as upstream/main. The host-side
  invariant holds: `stage.current` flips only inside dispatch, exactly when
  `bytes_written` resets, so every offset memcpy stays inside the generation
  acquired at `bytes_written==0`. The h_in release-by-alias (STAGING_FREE
  backed by the d-pool's H2D record) is sound.
- **Chunk pool**: produce-acquire before the re-zero (`drain_kick_and_swap`)
  and the consume-acquire in `acquire_inputs` reproduce #140's edge exactly;
  POOL_FILLED stays instance-folded to 0 and its record/wait interleaving
  (wait_k enqueued before record_k+1) is unchanged. Sync-path re-zero via
  `gpu_pool_at` is genuinely host-ordered (D2H_DONE poll ⇒ SLOT_DRAINED ⇒
  AGG_DONE ⇒ aggregate's pool read complete).
- **Aggregate facets**: no consumer reads through a facet ordered only by
  another facet's edges. The one cross-facet read — passthrough drain
  reading h_offsets under D2H_DONE — is ordered by same-stream earlier
  copies in the kick, so it's sound. CHUNK_INDEX_READY stays
  compressed-only on both record and poll sides; in passthrough runs it has
  zero records so the debug dead-edge check stays quiet. The
  release-even-on-dispatch-failure sites verify: if the release itself
  fails, the Error path still publishes the tail generation, and an
  unrecorded SLOT_DRAINED cannot park the next acquire (the event stays in
  its previously-signaled state) — the engine is already failing the write.
- **Tail / GEN_COUNTER vs #142**: `ordering.c` is untouched; only call
  sites moved. Threshold-advance-on-disabled-kick, exactly-once publish on
  every `sync_and_deliver` exit (including the early `sink->has_error`
  return), release_all-before-blocking-sync on both destroy paths
  (`tile_stream_gpu_destroy` and `compress_agg_destroy_shared`), and the
  `!gate_supported` depth-1 fallback keying are all preserved. Multiarray
  correctly never initializes the gate (NULL `gate_ord`), so its
  sync-everything destroy needs no release.

### `gpu_pool_at` sites — all seven justified

pool_epoch / partial-append (same compute stream, inside the swap's
produce-acquire), h_in memcpy (host acquire at fill start), sync-flush
re-zero + multiarray pool zero (host-completed sync drains), io_done peeks
in `d2h_deliver_kick` and `drain_d2h_for_array` (host bookkeeping ordered
by drain-before-rekick / sync-flush). The constraint is documented at the
declaration and each site carries its claim.

### Mutation checks (independent reproduction)

- Chunk-pool produce-acquire dropped (re-zero without the POOL_CONSUMED
  wait): `gpu_zstd_determinism` failed **5/5** reps.
- Tail gate never armed (`arm_tail_gate` enable forced 0):
  `gpu_page_aligned_tail_carry` failed **5/5** reps.
- Restored: **2/2** full `test_cross_validate` runs green.

### Validation (fresh builds, this review's worktree)

- `ctest -E "(s3)"`: **47/47** RelWithDebInfo, **47/47** Debug (Debug ran
  with all ordering/stream-declaration asserts live).
- `test_cross_validate` ×8: 8/8 green (all seven subtests each rep).
- `test_multiarray_gpu` ×4: 4/4. `two_streams` bench: zstd 47.7 s /
  3.93 GiB/s combined; none 28.0 s / 6.70 GiB/s — both clean.
- Warnings: PR and upstream/main RelWithDebInfo warning sets are
  byte-identical (17 pre-existing lines, all in `test_zarr_fs_sink.c` /
  `bench_util.c`) — zero new.
- Perf, 256cube_single zstd, base vs PR interleaved ×3 (same L40, discard
  sink): base 14.376/14.370/14.538 s, PR 14.315/14.284/14.328 s — PR
  **−0.8%** (faster, within noise); no rep near the 3% gate.

### Nits (no action required to ship)

1. `gpu_pool_release_all` (pool.c:119) forwards to
   `gpu_edge_release_all(ord)` — its scope is the whole ordering table, not
   the receiver pool. Today that's equivalent (one GEN_COUNTER edge; event
   edges never park), but the name promises per-pool scope; a one-line
   comment in pool.c would prevent surprise if a second gen-counter pool
   ever appears.
2. `flush_partial_append`'s comment (stream.flush.c:400) claims
   same-generation/same-stream ordering. That covers the epoch-aligned
   case; when a partial batch was just sync-flushed from the same slot, the
   actual guarantee is the host-completed drain (same class as the
   sync-flush re-zero above it). Suggest the host-ordered wording.
3. `drain_d2h_for_array` (multiarray/stream.gpu.c:69): "the sync above
   quiesced the slot" — the d2h sync doesn't cover the drain-stream record;
   what makes the io_done access safe is the sync-flush host-completed
   drains. Comment-only.
4. `gpu_pool_host_acquire_produce(handoff->tail, ...)` can't fail
   (consumed edge is GPU_EDGE_COUNT), so its `goto Error` is dead — fine as
   intent documentation.
5. Pre-existing, not this PR: the sync path double-records POOL_FILLED
   (flush_accumulate_epoch then flush_accumulated_sync) — harmless, maybe
   worth collapsing in step 4.

Boundary claim verified: exactly three direct edge ops outside
ordering/pool, all the documented LOD_DONE exemption — which is sound (a
second producer edge into the chunk pool can't be a pool slot without API
growth, and every LOD pool write happens inside the acquired produce
generation on compute).

PR #170, opened 2026-08-06
Benchmark report: trends and machine grouping

nclack:
The report showed one sweep at a time behind nine filter rows, so it could
not say whether performance was moving, or which machine a number came
from. report.py now writes two pages instead of one.

index.html is the new landing page: a card per machine (current value,
change against that machine's previous sweep, sparkline, failed runs), a
trend chart with one line per machine over sweep dates, a
scenario-by-machine table of the latest standings, and a table of the
largest changes since each machine's previous sweep. One filter row scopes
all of it, a table view mirrors the chart, and points link through to the
sweep they came from.

explore.html is the existing dashboard, unchanged in behaviour, now
addressable per sweep with ?file=. Both pages share light and dark themes.

bench/machines.toml declares which sweeps are the same machine. Neither the
sweep name nor the GPU model can decide that on its own: one box boots two
systems, a cluster hands out a different node per allocation, and storage
and network matter as much as the card. Entries match the sweep name or the
hostname, both with wildcards, and carry a description plus free-form
specs. Grouping stays a view — the overview splits a machine back into its
members, and says so when a comparison crosses two of them.

sweep.py --machine, or CHUCKY_MACHINE, names the results file and records
machine.name, so cluster runs no longer need renaming by hand.

report.py -o now names an output directory; pages.yml passes _site.

### DIFF
--- a/bench/machines.toml
+++ b/bench/machines.toml
@@ -1,0 +1,64 @@
+# Which sweep files belong to the same machine, and what to call it.
+#
+# A sweep is named after its results file, `<name>-<commit>-<date>.json`, which
+# sweep.py writes using --machine (or the hostname when that is not given). That
+# is not enough on its own: one box can boot two systems, and a cluster hands out
+# a different node every allocation. This file says which of those names are the
+# same machine.
+#
+# Each entry matches on the sweep name or on the hostname recorded inside the
+# file. Both accept shell-style wildcards and ignore case, and the first entry
+# that matches wins. A sweep matching nothing keeps its own name and the report
+# points it out, so a new machine gets described here on purpose rather than
+# quietly appearing as a stranger.
+#
+# `specs` is free-form: put whatever decides whether two runs are comparable.
+# Storage and network usually matter as much as the GPU.
+
+[[machine]]
+name = "livescreen"
+description = "RTX PRO 6000 workstation, two systems on one box"
+names = ["LiveScreen-1", "livescreen-kubuntu"]
+[machine.specs]
+gpu = "RTX PRO 6000 Blackwell Workstation Edition"
+
+[[machine]]
+name = "reef-l40"
+description = "CoreWeave L40 node, whichever one the scheduler hands out"
+names = ["reef-l40"]
+hosts = ["cw-us-e4a2-l40-*"]
+[machine.specs]
+gpu = "L40"
+storage = "shared /mnt/main0"
+
+[[machine]]
+name = "reef-h100"
+description = "CoreWeave H100 node, whichever one the scheduler hands out"
+names = ["reef-h100"]
+hosts = ["cw-us-e4a1-h100-*"]
+[machine.specs]
+gpu = "H100"
+storage = "shared /mnt/main0"
+
+[[machine]]
+name = "reef-h200"
+description = "CoreWeave H200 node, whichever one the scheduler hands out"
+names = ["reef-h200"]
+hosts = ["cw-us-e4a1-h200-*"]
+[machine.specs]
+gpu = "H200"
+storage = "shared /mnt/main0"
+
+[[machine]]
+name = "auk"
+description = "Laptop"
+names = ["auk"]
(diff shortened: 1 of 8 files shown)


nclack:
Notes for review — rationale, choices, and what was checked.

## Why a registry rather than inferring the machine

The overview needs to know which sweeps belong together. Nothing in the data
can decide it:

- **The sweep name** cannot: `sweep.py` names files after `platform.node()`, so
  `LiveScreen-1` and `livescreen-kubuntu` — one workstation, two systems — look
  like two machines, and every cluster allocation invents another one.
  `reef-l40-*.json` only reads correctly because the files were renamed by hand.
- **The GPU model** cannot: it does not separate two boxes with the same card,
  and it says nothing about the storage and network, which often decide the
  number as much as the GPU does.

That knowledge lives with the person running the sweeps, so it is written down
and reviewed like anything else. Matching on the hostname with wildcards
(`cw-us-e4a2-l40-*`) means a cluster sweep lands in the right machine even when
nobody renames the file, and `reef-h100` / `reef-h200` are pre-declared so the
first sweep from those partitions groups itself.

The risk of grouping is that it hides a real difference between the members, so
it is reversible and self-reporting: **Group** splits a machine back into its
sweep names, a comparison that crosses two members is labelled on both the card
and the movers table, and a group holding more than one GPU model says so.

## Presentation choices

- A plotted point is the **best** passing run for the selection; data type,
  chunk size and fill are searched, not averaged. The tooltip names the winning
  configuration and how many runs it beat.
- A sweep that ran nothing matching the filter leaves a gap, never a zero, and
  the empty view offers the nearest selections that do have data — asking for
  `blosc-lz4` on `gpu` suggests `backend cpu`, since blosc only ran on CPU.
- Changes under ±2% are called "no real change" rather than coloured.
- `retired_metrics` is honoured: a metric whose meaning changed between schema
  versions is dropped from comparisons instead of converted, and the page says
  so.
- Machine colours are assigned in order of first appearance, so a new machine
  never repaints the existing ones. The palette holds eight; past that,
  machines stay in the tables, leave the chart, and are named on the page.
  Series colours were validated for colour-vision deficiency and contrast
  rather than picked by eye (worst adjacent CVD ΔE 9.1 light / 8.4 dark).

## Checked

Both pages rendered headless, light and dark, with no console or page errors:
tooltip and crosshair, every metric including the nested `stalls.*` ones, every
codec, the scenario row click, empty-state recovery, the table view, all three
explorer views, `?file=` deep links, and theme persistence across pages.

Registry behaviour was checked directly: case-insensitive name matching, a
hostname glob catching an un-renamed cluster file
(`cw-us-e4a2-l40-88-2-abc1234-20260801.json` → `reef-l40`), a stamped
`machine.name` surviving an unparseable file name, unmatched sweeps staying
unmatched, and malformed registries (duplicate name, missing name, bad TOML)
stopping the build while an unknown key only warns.

Grouping the 21 existing files gives 4 machines instead of 5, `livescreen`
becoming one 10-sweep history across both installs.

## Previewing

```sh
uv run scripts/sweep/report.py --results-dir bench/results/ -o _site
python3 -m http.server -d _site      # or just open _site/index.html
```

Both pages embed their data, so `file://` works; d3 and pico still come from
CDNs. `machines.toml` is picked up from beside the results, or one level up.

bench/machines.toml
```diff
@@ -0,0 +1,53 @@
...
+# different node each time you ask for one.
+#
+# An entry matches a sweep by its name, or by the hostname stored in the file.
+# Both accept wildcards and ignore case, and the first matching entry wins. A
+# sweep that matches nothing keeps its own name, and the report marks it as not
+# described here.
+#
+# Fill in specs only where the sweep file does not already record something. It
+# records the GPU. It does not record the disk or the network, which matter just
+# as much when you compare two runs.
```
nclack:
```suggestion
# records the GPU. It does not record the disk or the network.
```

PR #172, opened 2026-08-06
Link libm only on platforms that have it

nclack:
#167 added `m` to stream_config's LINKS and to the aggregate library's public
dependencies. MSVC ships no m.lib — the math functions live in the C runtime —
so every Windows test link fails with LNK1181 and windows-tests has been red on
main since that merge.

Ask for libm through the same generator expression the file already uses for
its other conditional dependencies, so it is requested everywhere except
Windows.

Verified on Linux that the link line for test_aggregate_cpu is unchanged, -lm
still in the same position, and that the expression resolves to nothing under
CMAKE_SYSTEM_NAME=Windows.

### DIFF
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -16,8 +16,6 @@ function(add_src_lib TARGET)
     set_property(GLOBAL APPEND PROPERTY CHUCKY_INTERNAL_TARGETS ${TARGET})
 endfunction()
 
-# Unix keeps the math functions in their own library; MSVC has them in the C
-# runtime and ships no m.lib, so asking for one fails the link.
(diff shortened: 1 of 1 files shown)


src/CMakeLists.txt
```diff
@@ -16,6 +16,10 @@ function(add_src_lib TARGET)
     set_property(GLOBAL APPEND PROPERTY CHUCKY_INTERNAL_TARGETS ${TARGET})
 endfunction()
 
+# Unix keeps the math functions in their own library; MSVC has them in the C
+# runtime and ships no m.lib, so asking for one fails the link.
```
nclack:
```suggestion
```

PR #174, opened 2026-08-06
Dispatch when the staging buffer fills

### LINKED ISSUE #173: Every append starts its own transfer, however small
The writer starts a transfer and a kernel for every append, whatever its size. A
caller handing over one frame at a time gets one small transfer per frame instead
of the staging buffer being filled first.

The cost is per dispatch, not per byte. Measured on an L40 with the same 18.75 GiB
of data:

- handed over as 64 MiB blocks: 300 dispatches, 9.2 GiB/s
- handed over as 128 KiB frames: 153,600 dispatches, 3.0 GiB/s

Per call, the transfer takes 14.6 microseconds where the data itself accounts for
4.9, and the kernel takes 18.9 where the data accounts for 2.3. So about 26
microseconds of each append is fixed cost and about 7 moves data.

The pipeline then backs up behind it. The producer waits for a staging buffer
34,918 times, totalling 3.9 seconds, against 295 times totalling 1.5 seconds with
block-sized appends.

This is the normal path rather than an edge case, because feeding one frame at a
time is what the writer is for.

How much can be grouped depends on the geometry. The staging buffer is 128 MiB. In
the 256cube benchmark one epoch is 384 MiB, so appends could already be grouped up
to a full buffer within a single epoch. In the smallepoch benchmark one epoch is a
single 1 MiB chunk, so grouping within an epoch would cap a dispatch at 1 MiB —
still 2048 times better than a 512 byte frame, but well short of the buffer.

nclack:
Closes #173.

The writer started a transfer and a kernel for every append, however small. It now
fills the staging buffer first and hands over the whole thing.

A dispatch can then cover more than one epoch, so:

- The scatter runs once per epoch the buffer covers, into that epoch's own region
  of the pool. The transfer stays one call for the whole buffer.
- Epoch and batch accounting counts elements that have been dispatched rather than
  elements copied into staging. The append cursor still counts everything the
  caller handed over, and now runs ahead of the device between dispatches.
- Flush hands over a partly filled buffer, and counts every epoch that buffer
  completes rather than only a final partial one.
- A dispatch covers at most the room left in the batch, since the pool holds one
  batch of epochs. Multiscale covers at most one epoch, since its linear buffer
  holds one.
- Switching arrays hands over the departing array's staged data first. Staging is
  shared by every array in a stream.
- The scatter kernel loads whole 4 byte words covering the elements it was asked
  for and skips the elements before and after them, so a slice that starts partway
  into the buffer keeps the combined loads. It used to read a full 4 KB block past
  the end of its source; it now needs 4 bytes, which the staging buffer carries.

Producer memcpy costs 2.4x more, because each append used to write to the same
first part of the pinned buffer and now sweeps the whole thing. H2D and scatter
give back several times that.

Benchmarked on an L40 with the same data before and after. Frame sized appends:
256cube 3.0 to 10.9 GiB/s, smallepoch 0.02 to 3.0 GiB/s. Block sized appends are
unchanged to slightly better.


nclack:
Measurements are all from one L40 (sm_89, driver 580.126), RelWithDebInfo, u16
data, zstd.

Each row compares the same benchmark before and after the change. The two append
sizes within a benchmark are not comparable to each other. The benchmark fills one
block and repeats it, so a smaller block compresses better and moves work out of
the compressor.

| benchmark | append size | throughput before | after | dispatches before | after |
|---|---|---|---|---|---|
| 256cube | 64 MiB | 9.4 GiB/s | 9.3 GiB/s | 300 | 150 |
| 256cube | 128 KiB | 3.0 GiB/s | **10.8 GiB/s** | 153,600 | 150 |
| smallepoch | 64 MiB | 3.9 GiB/s | 4.7 GiB/s | 977 | 8 |
| smallepoch | 512 B | 0.02 GiB/s | **3.0 GiB/s** | 2,000,000 | 8 |

Detail for the frame sized appends:

| | 256cube before | after | smallepoch before | after |
|---|---|---|---|---|
| wall | 6.34 s | 1.73 s | 38.3 s | 0.32 s |
| scatter | 6.4 GiB/s | 77.3 GiB/s | 0.1 GiB/s | 91.2 GiB/s |
| waits for a staging buffer | 35,146 (3.9 s) | 19 (0.15 s) | 147,173 (15.7 s) | none |
| append time p99.9 | 0.141 ms | 0.028 ms | 0.126 ms | 0.0011 ms |

## Epoch and batch accounting

Epoch and batch accounting depended on the append cursor, and the cursor advanced
inside the dispatch. That was correct only because each pass through the append
loop stopped at an epoch boundary. Once a dispatch waits for a full buffer, it can
carry data past a boundary while that data is still in staging. The level of detail
work would then run over a pool region nothing had written, and at a batch boundary
that region would be compressed and stored. The result is wrong data rather than a
stall.

The cursor now counts elements the caller has handed over. Accounting counts
elements that have been dispatched. The two are equal only between dispatches.

## Scatter kernel

The kernel read a full 4 KB block past the end of its source and left the caller to
provide that room. This worked while the source was always the whole page aligned
staging buffer, and stops working for a slice inside it. The kernel also needed a
4 byte aligned start so it could combine loads, which a slice starting partway
through an epoch cannot give it.

It now loads whole 4 byte words covering the elements it was asked for and skips
the elements before and after them. Every slice keeps the combined loads, and the
source needs 4 bytes of room past its end. Some threads at each boundary do
nothing; the benchmarks show no cost.

## Next steps

Step two of #173 gives the scatter the shape of a batch, so one launch covers every
epoch instead of one launch per epoch. It is not included here. Scatter now takes
243 ms of a 1.7 s 256cube run and 10 ms of a 0.32 s smallepoch run, at 77 and
91 GiB/s, so the 128 launches per buffer are not a bottleneck.

Multiscale streams still group appends only up to one epoch, because the buffer they
copy into holds one epoch. Sizing that buffer to a batch would let them group like
the single scale streams. A multiscale benchmark would show whether that is worth
doing.

src/gpu/transpose.h
```diff
@@ -5,6 +5,11 @@
 #include <cuda_runtime.h>
 #include <stdint.h>
 
+// transpose reads whole 4-byte words overlapping the elements it is given, so a
+// source buffer needs this much room past the last element it will be asked
+// for.
```
nclack:
```suggestion
// transpose reads whole 4-byte words overlapping the elements it is given, so a
// source buffer needs this much room past the last element
```

nclack:
Review findings, worked through.

## Fixed

Three defects, all reachable only after an earlier failure.

A dispatch that fails partway leaves the epochs it transferred uncounted, and
nothing can un-enqueue them. Later work would place data using a stale epoch
position. The engine now records that and stops taking data, which also closes
the case where an array switch failed and left staged bytes to be sliced with
another array's epoch size.

A failed kick leaves the batch full without swapping the slot. An append then
computed the room left in the batch as `epochs_per_batch - accumulated`, which
underflowed, and the scatter wrote past the slot's last epoch. Appends now stop
in that state, and counting an epoch into a full batch is refused rather than
running past `batch_active_masks` as well.

The scatter reads into the 4 bytes past the staging buffer and discards them, but
no transfer wrote there, so it read as uninitialized to a checker. Zeroed once at
init.

Two more, for robustness rather than a live bug: the multiscale copy now asserts
that the staged bytes stay inside one epoch, which is a rule the caller keeps and
this file cannot see; and the zero check on `epoch_elements` moved into
`stream_dispatch_staged`, which had three call sites and one guard.

## Reported behaviour

Producer memcpy is 2.4x slower: 584 to 1398 ms at the same call count, 32.1 to
13.4 GiB/s. Each append used to write to the same first 128 KiB of pinned memory
and stay in cache, and now sweeps the whole buffer. That is 813 ms added to the
producer thread against 1553 ms taken off H2D and 2695 ms off scatter.

The slow appends that remain are slower: the ones over 1 ms moved from around
3.2-3.6 ms to 5.6-7.1 ms, because a dispatch-carrying append enqueues 128 MiB and
up to K launches instead of 128 KiB. Fewer of them, and the longest append fell
from 60.1 to 45.9 ms.

## Tests

`transpose_unaligned_source` ran a single block, so it never exercised the load
offset for any block past the first. Its shape now spans 3 blocks at 1 byte per
element and 6 at 2.

`cross_validate_dispatch_spans_epochs` is new: 9 element epochs against a 4096
byte staging buffer, so every dispatch after the first starts partway into an
epoch, covers several, and lands slices on odd elements. It compares GPU output
against CPU output byte for byte.

`batch_mid_batch_after_dispatch` pins a batch that is counted but unkicked, which
no test covered after the change.

`test_batch`'s first case was named for the staging buffer, but what holds the
data there is the batch cap. Renamed, and two comments that said the same thing
corrected.

## Other

`scatter_destination` took a pool handle and a slot index, which put pool
arithmetic inside a stage against the direction in dev/gpu-orchestration.md. It
carries the first epoch's view instead.

Comments that restated their function are gone. `dispatch_at_bytes` reads as a
verb but returns a threshold, so it is `bytes_before_dispatch`; `lead` is
`lead_elements`. `batch_accumulated` and `peak_pending_bytes` now say what they
mean after the change. `staging_slot.h_in` claimed write-combined memory, which it
has never been.

Full suite green on the L40, and the benchmark numbers are unchanged by any of
this.

nclack:
Second review, of the commit that answered the first one. Two of its findings were
regressions I had just introduced.

## Fixed

A stream that stopped taking data also stopped closing its output. `stream_flush_body`
began with the staged dispatch and returned its error, so the shard indexes, the
array shape and the queued writes were all skipped — full shards were on disk with
nothing to read them by. Flush is now split: the part that places more data is
skipped once the position is unknown, and the part that closes the sink always runs.
The failure is still what gets returned. This also means the closing work now happens
after any flush error, not only this one, which is the rule
tests/test_flush_error_drain_cpu.c holds the CPU backend to.

One array's failure stopped every other array in a multiarray stream from being
finalized, because the flag is engine-wide and `flush_impl` bailed before its loop.
The loop now runs to the end and reports the first failure.

A batch that was full but never kicked could not be flushed at all: the guard added
last round refused to count the partial epoch, and the only thing that resets the
count is the kick that never happened. Flush now kicks a full batch before counting
another epoch into it.

The pad memset was aimed at the wrong bytes. The scatter's overread is relative to
the last element transferred, not to the end of the buffer, and since the buffer is
page-aligned the read never leaves it — so the 4 bytes past the end were never read,
while the few bytes between the last element and the next word boundary were. The
whole buffer is zeroed once at init, which covers every case.

`scatter_destination` now carries the number of regions the caller holds, so the
stage asserts its own bound rather than trusting an invariant kept in another file.

## Reporting

The sweep schema is at version 3, retiring `peak_pending_mib` and the backpressure
pair. Sampling moved from once per epoch to once per staging buffer, which on
smallepoch is a drop from about 65,000 samples to 8, so the peak reads lower for a
reason that has nothing to do with the machine. Without the bump the overview would
have drawn that as an improvement.

## Documents

dev/gpu-orchestration.md gains a paragraph on the four sizes of work — append,
dispatch, epoch, batch — which is now the subsystem's main performance mechanism and
was not written down anywhere. Item 1 of "What's next" is marked done, since all five
of its measurement problems were fixed in #163 through #167. The note declining CUDA
Graphs said to revisit if dispatch overhead ever proved a real cost; it did, and the
answer was to dispatch less often, so the note says that now. #173 is in "Related",
and the `gpu_pool_at` caller count was two low.

docs/design.md's `buffer_capacity_bytes` row now says it is rounded up to a page and
that it sets the dispatch size, so a caller tuning for latency can see the trade.

## Tests

`batch_failed_delivery_still_closes_sink` is new and reaches the failure path all
three fixes live in: a shard too small for a batch makes delivery fail at the third
batch's kick, the append reports it, the next append is refused, and the flush
reports the failure while still writing the array shape.

`batch_mid_batch_after_dispatch` passed on main as well, so it could not catch a
batching regression. It now also checks that the 16 epochs crossed in one transfer.

Full suite green on the L40, CPU-only configuration green, benchmarks unchanged.

nclack:
Third review. Its two must-fix findings say the previous round's fix went the wrong
way, and they are right.

Closing the sink unconditionally traded one problem for two worse ones. Nothing joins
the delivery worker except `schedule_drain_kicked`, which sat inside the part that got
skipped — so `finalize_shards` could run on the same `shard_state` the worker was
still writing, and both threads could call `update_append` at once. And the array
shape comes from the append cursor, which counts data that never reached the sink, so
a failed flush rewrote `zarr.json` to claim frames that are not there. A reader gets a
full-size array with a fill-value tail and no way to tell.

The rule is now the other way round. Joining the delivery worker and draining queued
writes happen on every path, because they are cleanup. Finalizing a shard and writing
the array shape happen only when everything before them succeeded, because both of
them claim the output is complete. After a failure the output is left short rather
than made to look whole.

The pre-kick added last round for a full-but-unkicked batch is gone. It counted the
same partial epoch twice, the second time over a pool region that still held the batch
just delivered, so a retried flush appended an epoch of stale data. That state is
terminal instead: a failed kick sets the same flag a failed dispatch does, since
nothing can say which of the batch's epochs reached the sink.

The new test asserts the opposite of what it did last round: after a failed delivery
the array shape must *not* be written. `update_append` count is 0 where I previously
required it to be non-zero.

Also from this round: the `gpu_pool_at` paragraph in dev/gpu-orchestration.md said
three of its callers are pool zeroing, and two are. The `buffer_capacity_bytes` row in
docs/design.md now says a dispatch never exceeds one batch's remaining bytes, so
raising the buffer past that does nothing — which is what someone reading that row for
latency reasons needs to know. And the duplicated `epoch_elements == 0` check is one
check again.

One correction to my earlier description: I said this makes the GPU path follow the
rule tests/test_flush_error_drain_cpu.c holds the CPU backend to. That test only
requires the sink be drained before returning an error, which is the part I kept. It
is not a precedent for finalizing shards or rewriting the shape after a failure, and
the CPU backend does neither.

Full suite green on the L40, benchmarks unchanged. CI is running again now that the
Actions outage has cleared.

nclack:
Fourth review. One of its two must-fix findings is mine; the other turned out to be
main's behaviour, and is now issue #175.

## Fixed

The flag that stops a stream taking data was on the engine, shared by every array of
a multiarray stream, so one array's failure made all the others skip finalizing their
shards and writing their shape. Their data is complete — a switch flushes and drains
the departing array, and multiarray drains after every kick — so there was nothing
wrong with it to protect anyone from. The flag is per array now, on the context that
owns the cursor it invalidates. A dispatch failure also lets go of the staged bytes,
since staging is shared and those bytes are never reaching that array's pool.

Two failure paths did not set the flag: a failed `schedule_drain_kicked` and a failed
`schedule_flush_partial_append`. The second mattered — its predecessor has already
reset the batch, so destroy's auto-flush would count the same partial epoch a second
time, deliver it, and then finalize and publish over it. Both latch now.

## Not fixed, filed as #175

The review's second must-fix is that skipping the shape write after a failure leaves
the recorded extent behind whatever the last periodic update wrote, so delivered and
finalized shards can be invisible to a reader. That is real, but it is what `main`
does: every one of its early returns in `stream_flush_body` skipped `update_append`
too. This PR changed it for one commit and changed it back.

The fix is not a gate, it is the derivation. The shape comes from the append cursor,
which counts data still in staging, so it can only be written when everything
arrived. The delivered extent — which the periodic update already computes — is
correct either way and could be written unconditionally. That is #175, with the
approach in a comment. It touches the CPU backend the same way, so it does not belong
in this PR.

## Also

The test is `batch_failed_delivery_writes_no_shape`, which is what it checks; the old
name claimed it closed the sink and it had stopped asserting anything of the sort. It
now also checks the sink was drained, which is the other half of the rule. The finalize
half of that gate still has no coverage: with a shard too small for even the first
delivery, `epoch_in_shard` never advances, so finalize is a no-op either way. A sink
sized to pass one batch and fail the next would cover it.

The `buffer_capacity_bytes` row said raising it past one batch has no effect; it has
no further effect on grouping, but it does still cost pinned memory.

Full suite green on the L40. CI is green on the previous commit across cpu, macOS,
Windows and both CUDA versions.

nclack:
Two tests written to pin the per-array fix, and both failed. Both were right to.

Switching away from a failed array returned an error, so the stream got stuck on the
broken one and every array after it was starved. Two places did it: the dispatch of
the departing array's staged bytes, and the flush of its accumulated batch. Neither is
the incoming array's problem — the failure belongs to the array being left, which
reports it when it is flushed. This is the same mistake as the engine-wide flag, one
level down, and it made that fix useless in practice: the other arrays were spared the
flag and then never got any data.

`test_one_array_failure_spares_the_others` covers it. One of three arrays gets a sink
that refuses every write; the flush reports the failure and the other two still
finalize their shards.

The second test aimed at the other half of the gate — a shard holding delivered data
that does not get an index because a later batch failed — and I could not construct it.
Delivery finalizes a shard itself once it fills, so the count says nothing, and in this
geometry the delivered batch filled its shard exactly, leaving no partial one to
observe. I dropped the test rather than keep tuning it into a pass. So: the shape half
of that gate is covered, the index half is not, and I would rather say that than ship a
test that passes for a reason it does not claim.

To make the failure land at a chosen point rather than on the first write, the test
sink takes a byte budget (`fail_writes_after_bytes`). Both new tests use it.

Benchmarks unchanged. Full suite green on the L40, and CI is green so far on this
commit.

nclack:
Last review round. Its one must-fix was an assertion of mine that could not fail, which
is the thing I have been holding other tests to.

`shard_sink_pending_bytes(&css.base) == 0` is vacuous: the test sink never installs
`pending_bytes`, so it returns 0 whatever happens, and it installs no fence hooks either,
so `shard_sink_drain` is a no-op for it. Put the drain back behind the success gate and
the suite would still pass. Removed, along with the claim that the test covers that half
of the rule. The CPU side pins it observably through the real filesystem pool; the GPU
side does not, and saying so is better than an assertion that only looks like it does.

Two more places treated a departing array's failure as the incoming array's problem, so
the fix from the previous commit still did not hold in general. The mid-epoch rejection
fires before the failure is considered, and a failed array can never finish its epoch, so
one bad array could still strand the rest permanently. And the switch kicked a failed
array's accumulated batch, which the flush deliberately refuses to do — nothing can say
which of that batch's epochs reached the sink, so kicking it again re-delivers chunks.
Both are now behind one check on the departing array's own state.

The multiarray test only exercised one of the two changes: with the default batch size,
array 1's sink was never touched by array 1's own append, so the failure was latched
during the switch rather than before it. Array 1 now takes one epoch per batch, which
puts the kick and the sink failure inside its own append, and the test asserts that
directly — array 0 succeeds, array 1 errors, array 2 succeeds.

Also: the byte budget in the test sink charges only accepted bytes now, and reads as a
predicate instead of mutating while it answers. The duplicated `append_failed` check in
the flush is one check.

Full suite green on the L40; CI was green on the previous commit across all six jobs.

That is the end of the review passes. Two things are knowingly left: the finalize half of
the flush gate has no test, for the reason in the previous comment, and #175 is the
metadata extent, which is `main`'s behaviour and touches the CPU backend the same way.

nclack:
Measured write-combined staging memory as a fix for the producer memcpy
regression. Recommendation is to leave it alone.

One L40, three repetitions, base and patched interleaved within each rep, clocks
pinned at 2490 MHz throughout. Two build trees from the same source differing
only by the `cuMemHostAlloc` flag. Nothing on the host reads `h_in` — checked
every use of the staging payload first — so the change is safe, just not
worthwhile.

| benchmark | append size | copy path | memcpy base | memcpy write-combined |
|---|---|---|---|---|
| 256cube | 128 KiB | single thread | 1331 ms, 14.1 GiB/s | **812 ms, 23.1 GiB/s** |
| 256cube | 64 MiB | thread pool | 410 ms, 46.2 GiB/s | 461 ms, 41.9 GiB/s |
| smallepoch | 512 B | single thread | 47.6 ms, 20.0 GiB/s | 63.8 ms, 14.9 GiB/s |
| smallepoch | 64 MiB | thread pool | 17.7 ms, 54.0 GiB/s | **31.4 ms, 30.6 GiB/s** |

It does what it was supposed to on the one case it was aimed at: the 128 KiB
single-threaded copy drops 39%, and the gap is about thirteen times the
run-to-run spread, so that is real. It recovers roughly two thirds of the time
lost since the change, though only half the bandwidth — the figure before was
584 ms and 32.1 GiB/s.

It does not move wall time. 1.716 versus 1.711 seconds with overlapping ranges,
because compression runs 1.59 seconds and sets the pace. The saved producer time
turns into waiting: the staging-free stall grows from about 180 ms to about 720
ms.

And it costs the path both benchmarks use by default, where the copy goes to the
thread pool. smallepoch loses 78% on the stage and 6.3% of end-to-end throughput,
with cleanly separated ranges. The likely reason is that a 64 MiB copy already
uses non-temporal stores, so there is no read-for-ownership left to avoid and
write-combining only adds contention for the write-combining buffers.

Device transfer is unaffected: H2D sits between 24.96 and 25.12 GiB/s across all
eight configurations, so reading the buffer by DMA is no slower.

The flag is fixed at allocation and one staging buffer serves every append size,
so it cannot be taken selectively. Since it helps a case whose win is currently
absorbed by the compressor and hurts the common case, it is not worth taking. The
useful follow-up is in `ingest_copy` instead, which is #183.

One confusing result: on smallepoch with 512-byte appends the memcpy stage gets 34% 
slower while wall time gets 7.3% faster, both cleanly separated. Plausibly the stores 
stop evicting the producer's own working set and the stage timer only sees the issue 
cost, but that is a guess.

Full suite passes 50/50 with the flag on.

PR #179, opened 2026-08-08
Fetch report data instead of embedding

### LINKED ISSUE #171: Report pages embed every sweep, so they grow linearly with bench/results
Both report pages embed every sweep at generation time, so they grow linearly
with `bench/results/`. That is fine today and will not stay fine.

Measured at 22 sweeps / 5,111 runs (the tree after #169, with the two-page
report from #170):

| page | raw | gzipped | per sweep |
|---|---|---|---|
| `index.html` | 2,267 KiB | 228 KiB | ~101 KiB |
| `explore.html` | 5,121 KiB | 731 KiB | ~230 KiB |

Pages serves gzip, so what crosses the wire today is 228 KiB and 731 KiB —
not a problem. The problems are further out:

- **It is linear.** At 100 sweeps: 10 MB and 23 MB raw, roughly 1 MB and
  3.3 MB gzipped. At 200: 20 MB and 46 MB.
- **The explorer downloads everything to show one sweep.** Per-stage timing is
  44% of its payload (2,254 KiB) and per-run stalls another 10%, and all of it
  is parsed on load even though the page only ever renders one sweep at a time.
- **Every regeneration invalidates the whole file.** Adding one sweep re-sends
  all of them; nothing can be cached per sweep.
- Raw size, not transfer size, is what the browser parses and holds.

Options, roughly in order of payoff:

1. **Split the explorer into per-sweep JSON, fetched on demand.** It drops to a
   small index plus one ~230 KiB fetch per sweep viewed, and adding a sweep
   invalidates only its own file. The cost is real: the page stops working
   from `file://`, which today is how a generated report gets previewed
   (`scp` the two files and open them). Would need a local server, or a
   `--self-contained` flag for that case.
2. **Encode runs columnar in the overview.** Its 101 KiB per sweep is mostly
   the same ~20 short keys repeated per run, ~5,000 times. Sharing the keys, or
   interning the repeated scenario/codec/dtype strings, should cut it several
   fold with no change to how the page is served or previewed.
3. **Bound what the overview loads** — last N sweeps per machine, with the rest
   reachable in the explorer. Simple, but it gives up exactly the long history
   the page was built to show, so it is a last resort.
4. **Do nothing yet, but notice.** Fail or warn in the Pages build when a page
   crosses a threshold (say 1 MB gzipped), so this gets attention on the commit
   that causes it rather than months later.

Suggest (2) and (4) now, since neither costs the `file://` workflow, and (1)
when the explorer becomes the thing people actually wait on.

Numbers reproduce with:

```sh
uv run scripts/sweep/report.py --results-dir bench/results/ -o _site
gzip -9 -c _site/index.html | wc -c
```

nclack:
Both report pages embedded every sweep, so they grew linearly with
`bench/results/` and the explorer downloaded all 22 sweeps to show one.

The pages are now code only. Their data is written beside them and fetched at
load: `data/overview.json` for the overview, and one file per sweep under
`data/sweeps/` that the explorer fetches when that sweep is opened. Inside those
files the runs are stored as columns instead of one object per run, with all the
text in a shared table (`columnar.py`), which drops the repeated key names and
run ids that were most of the payload. Floats are cut to four significant
figures, finer than any page prints and about what the benchmarks resolve.

At 22 sweeps / 5,111 runs:

| | before | after |
|---|---|---|
| overview first load | 2,268 KiB raw / 216 KiB gzip | 537 KiB raw / 115 KiB gzip |
| explorer first load | 5,121 KiB raw / 702 KiB gzip | 228 KiB raw / 48 KiB gzip |

Adding a sweep now leaves the other sweep files byte-identical, so anything
holding a cached copy can revalidate them instead of re-downloading.

The pages must be served over http, which is what `--serve` is for:

```sh
uv run scripts/sweep/report.py --results-dir bench/results/ -o _site --serve
```

Opening `_site/index.html` from disk no longer works, because the browser
refuses the fetches. That was the tradeoff accepted for splitting the data out.

`report.py` unpacks each sweep it just packed and refuses to write one the pages
would read back differently. CI no longer builds on results-only commits.

Closes #171.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,9 +1,12 @@
 name: CI
 
+# A sweep result is data, not code: nothing it can change is what CI builds.
 on:
   push:
     branches: [main]
+    paths-ignore: ['bench/results/**']
   pull_request:
+    paths-ignore: ['bench/results/**']
 
 concurrency:
   group: ci-${{ github.ref }}
--- a/.gitignore
+++ b/.gitignore
@@ -14,3 +14,4 @@ __pycache__/
 vcpkg_installed/
 .pre-commit-config.yaml
 docker-compose.override.yml
+_site
--- a/scripts/sweep/README.md
+++ b/scripts/sweep/README.md
@@ -14,10 +14,11 @@ sweep in `explore.html`.
 ## Generating the site
 
 ```sh
-uv run scripts/sweep/report.py --results-dir bench/results/ -o _site
-python3 -m http.server -d _site      # then open http://localhost:8000/
+uv run scripts/sweep/report.py --results-dir bench/results/ -o _site --serve
 ```
 
+That writes the site and serves it at http://127.0.0.1:8000/index.html. Pass a
+port to `--serve` to use another one, or drop the flag to only write the files.
 Run the command again after you add or change a results file.
 
 The pages are code only. Their data is written beside them and fetched at load:
@@ -31,10 +32,14 @@ The pages are code only. Their data is written beside them and fetched at load:
 So the explorer downloads one sweep instead of all of them, and adding a sweep
 leaves the other sweep files untouched for anything holding a cached copy. The
 cost is that the pages have to be served over http — opening `_site/index.html`
-from disk gets you an empty page, because the browser refuses the fetches.
+from disk gets you an empty page, because the browser refuses the fetches. That
+is what `--serve` is for.
 
 Inside those files the runs are stored as columns rather than one object per
 run, with the text in a shared table (`columnar.py`). The pages undo it on load.
+Floats are cut to four significant figures, which is finer than any page prints
+and about what the benchmarks resolve. `report.py` unpacks every sweep it packs
+and refuses to write one the pages would read back differently.
 
 `report.py` looks for `bench/machines.toml` next to the results directory, then
 one level up. Use `--machines` to point somewhere else.
--- a/scripts/sweep/README.md
+++ b/scripts/sweep/README.md
@@ -25,6 +25,7 @@ The pages are code only. Their data is written beside them and fetched at load:
 
 | file | holds |
 |---|---|
+| `decode.js` | unpacks the columns, imported by both pages |
 | `data/overview.json` | every sweep, trimmed, for `index.html` |
 | `data/sweeps.json` | the sweep list `explore.html` offers |
 | `data/sweeps/<result>.json` | one sweep in full, fetched when it is opened |
@@ -36,10 +37,11 @@ from disk gets you an empty page, because the browser refuses the fetches. That
 is what `--serve` is for.
 
 Inside those files the runs are stored as columns rather than one object per
-run, with the text in a shared table (`columnar.py`). The pages undo it on load.
-Floats are cut to four significant figures, which is finer than any page prints
-and about what the benchmarks resolve. `report.py` unpacks every sweep it packs
-and refuses to write one the pages would read back differently.
+run, with the text in a shared table. `columnar.py` writes that and `decode.js`
+reads it. Floats are cut to four significant figures, which is finer than any
+page prints and about what the benchmarks resolve. `columnar.py` unpacks every
+sweep it packs and refuses to hand back one that does not match, so a broken
+encoding stops the build rather than reaching a page.
 
 `report.py` looks for `bench/machines.toml` next to the results directory, then
 one level up. Use `--machines` to point somewhere else.
--- a/scripts/sweep/README.md
+++ b/scripts/sweep/README.md
@@ -15,11 +15,26 @@ sweep in `explore.html`.
 
 ```sh
 uv run scripts/sweep/report.py --results-dir bench/results/ -o _site
-python3 -m http.server -d _site      # or open _site/index.html
+python3 -m http.server -d _site      # then open http://localhost:8000/
 ```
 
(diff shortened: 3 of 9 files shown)


nclack:
Notes on what was tried and rejected, and one bug found along the way.

### An existing table format would have been worse

Measured against the same data, on the wire (Pages gzips `.html`/`.json`, and
does serve `accept-ranges: bytes`, so range-served SQLite was on the table):

| candidate | wire (gzip) | raw | plus runtime |
|---|---|---|---|
| columnar JSON, overview fields | **90 KiB** | 425 KiB | — |
| Parquet f32 + snappy | 122 KiB | 142 KiB | 16 KiB (hyparquet) |
| Parquet f32 + zstd | 128 KiB | 133 KiB | 16 KiB |
| columnar JSON, incl. stages | **354 KiB** | 1,790 KiB | — |
| SQLite, incl. stages | 850 KiB | 2,600 KiB | 313–495 KiB wasm |
| DuckDB-wasm | — | — | 6,878 KiB wasm |

Parquet is genuinely compact on raw bytes but loses on the wire, because it
arrives pre-compressed and gzip cannot improve it while compact JSON gzips 4.7x.
SQLite loses outright: rows repeat every string, indexes and page padding add
bytes, and `REAL` is a fixed 8-byte double so trimming digits saves nothing.
Range requests would also defeat gzip, and the overview does a full scan anyway.

Hand-rolled string interning was also mostly not worth it — gzip is already a
string interner. Per-column dictionaries bought 224 KiB raw but only 7 KiB
gzipped. The shared table in this PR is different: it dedupes run ids *across*
sweeps (568 distinct out of 5,111), which no per-column scheme reaches.

### Four significant figures was rejected

It measured well — another ~30 KiB gzipped off the overview — but rendering the
pages before and after showed it changing displayed numbers, including flipping
a mover's direction arrow from ▼ to ▲. The movers panel reports differences
smaller than four digits resolve, and values near 1.0 cross `fmt()`'s formatting
boundary when rounded. Settled on six digits, which is what `sweep.py` already
records, so it is a no-op today and only guards against future full-precision
output.

### Pre-existing bug: movers silently drops comparisons

Not fixed here, since it changes what the page reports and this PR is meant to
be a pure encoding change.

`moversFor` matches runs between two sweeps by `run.id`. That id carries a
suffix whose convention has drifted — `__fs`, `__io`, `__s3__10gbps`, or nothing
— and the suffix does not track the run's `sink` field. 144 of 5,111 runs use a
convention their neighbours do not. So the same configuration measured in two
sweeps written at different times fails to match, and the comparison is dropped
without any indication.

Keying on the config tuple instead (`scenario`, `codec`, `fill`, `backend`,
`dtype`, `chunk_bytes`, `sink`) makes many more rows match, and materially
changes the panel — on the current data the top mover for one machine goes from
`256cube_single −6.61%` to `orca2_single +128%`. Worth its own change; happy to
open an issue.

### How this was verified

No tests exist for the report pipeline, so correctness rests on two things.

`report.py` now unpacks every sweep it packs and refuses to write one that does
not match (`check_roundtrip`). That check was confirmed to actually trip, on a
swapped number, a dropped nested stall, a corrupted string table, and a shifted
text index.

Separately, the site was generated from `main` and from this branch, both served
over http and driven with a headless browser: `innerText`, all SVG label text,
and SVG element counts compared across five scenarios — the overview default
view, the overview on a second metric, the explorer default sweep, the explorer
after switching sweeps, and an `?file=` deep link. All five are identical with
zero console errors. Network capture confirms the explorer fetches two files on
load, one more per newly opened sweep, and none when revisiting.

nclack:
**Correction to the comment above: the "movers bug" is not real. I was wrong, and
the reasoning that led to it was wrong too.**

I claimed `run.id` carries a drifted suffix so the movers panel silently drops
comparisons, and that keying on the config tuple would fix it. Checking properly:

Under the default filter, id-matching and config-matching produce **identical**
results — 100 comparable rows either way, across all four machines. Nothing is
dropped.

The `__io` suffix is not a naming drift. It marks a genuinely different run. In
`LiveScreen-1-a799de5-20260320.json` these two coexist:

| field | `…u16__32K` | `…u16__32K__io` |
|---|---|---|
| scenario, codec, fill, backend, dtype, chunk_bytes, sink | identical | identical |
| `throughput_in_gibs` | 10.7331 | 6.33761 |
| `flush_s` | 1e-07 | 0.18867 |

Same config fields, different measurement. The id is the only thing telling them
apart. So `run.id` is the correct key, and the config tuple I proposed is the
defect — it merges 24–48 distinct runs per sweep in four sweeps
(`LiveScreen-1-a799de5`, `auk-d62fd7e`, `oreb-a799de5`, `oreb-ba38174`), keeping
whichever lands last. That merging is what produced the `256cube_single −6.61%`
→ `orca2_single +128%` shift I misread as a fix.

The shipped code keeps `run.id`, so the PR was never affected — only my claim
about it. No follow-up issue is needed.

**This also clears four significant figures**, which is now what the PR does. My
stated reason for backing off — that it flipped a mover's arrow — came from a run
that had the config-tuple change in it as well. With `run.id` kept, rounding to
four digits changes only the last printed digit of some percentages
(`−4.03%`→`−4.04%`), plus `1.000x`→`1.00x` where a compression ratio rounds to
exactly 1 and `fmt()` switches branches. Across all five checked scenarios there
are 53 word-level differences, no arrow flips, no row or ordering changes, and
the explorer is identical — including on a sweep containing the `__io` twins.

PR #180, opened 2026-08-08
Query nvcomp for compressed chunk alignment

### LINKED ISSUE #162: GPU compressed-output chunks are not 4-byte aligned as nvcomp requires
nvcomp requires every buffer passed to its batched compress calls to start at an address that is a multiple of 4. Both codecs we use document this: `nvcompZstdRequiredCompressionAlignment` and `nvcompLZ4RequiredCompressionAlignment` are 4, and the requirement covers input, output, and temporary buffers.

The GPU backend lays the per-chunk compressed-output buffers out back to back in one allocation, spaced by the codec's maximum output size per chunk: `fill_ptrs_kernel` computes `comp_base + i * comp_stride` (src/gpu/compress.cu:74 at 46fa2e8), and the stride passed in is `c->max_output_size` (src/gpu/compress.cu:342), taken directly from `nvcompBatched*CompressGetMaxOutputChunkSize`. For zstd that value is not a multiple of 4 — for example 65547 for 64 KiB chunks — so every compressed-output chunk after the first starts at a misaligned address.

nvcomp 5.2 accepts this today, but it violates the documented requirement, so a future nvcomp release could reject the call or corrupt output without warning.

nclack:
nvcomp reports the alignment each buffer needs through `GetRequiredAlignments`. The GPU backend assumed a constant instead — the value nvcomp advertises as the strictest across input, output and temporary buffers. For compressed chunks the requirement is reported separately and is currently 1, so nothing was actually misaligned, but the assumption would not track a future release that tightens it.

`codec_max_output_size` now asks nvcomp what a compressed chunk needs and rounds its bound to that. On nvcomp 5.2 this is a no-op: the stride stays at nvcomp's raw bound, with no padding. The pointer-fill stride, the compressed-pool allocation, the aggregate read stride and the device-memory estimate all derive from that one value.

Cleanups the investigation turned up:

- `codec_init` no longer repeats the max-output and temp-size queries that `codec_max_output_size` and `codec_temp_bytes` already implement
- the GPU support check uses the shared codec predicate instead of a hand-written blosc test
- the write-only `codec::alignment` field is gone

The new test asserts the bound covers what nvcomp says it can emit and is a multiple of the alignment nvcomp reports for a compressed chunk.

Closes #162


nclack:
Validation, on an L40 (sm_89) with nvcomp 5.2 and CUDA 13.1:

- clean configure + build, warning-free
- `ctest -E "(s3)"` — 50/50 passed
- re-ran `test-test_compress` verbosely to confirm the new case executes rather than being skipped: `PASS: max_output_size_alignment`

The misaligned values come from an earlier microbench on the same hardware. zstd `GetMaxOutputChunkSize` returns 16395, 65547, 262166, 524322, 1048634 and 2097258 for 16K through 2M chunks; none are multiples of 4, and only 128K (131088) happens to land aligned. LZ4 returns 8-aligned values at every size, so this was zstd-only in practice. nvcomp 5.2 accepts the misaligned pointers and produces correct output — the new test fails against those raw values, not against anything observable at runtime today.

Two things deliberately left alone. `codec_init` still makes its own `GetTempSizeAsync` calls rather than going through `codec_temp_bytes`; that function returns 0 both for `CODEC_NONE` and on failure, so using it would turn an init-time error into a compress-time one. And the bound is rounded to the documented 4 rather than 16 or 32 — the headers hint larger alignments may help, but the microbench measured no effect.

Independent of #174 and #177: neither touches `compress.cu`, `compress.h` or `test_compress.c`, and neither adds or removes a line mentioning `max_output_size`.

PR #184, opened 2026-08-08
Record whose bytes staging holds

nclack:
`struct staging_state` is one buffer shared by every array of a multiarray
stream. Its bytes belong to one array at a time, and nothing recorded which;
the rule lived in a comment upheld by hand at three call sites.

Staging now names its owner. It gains `owner` and `first_element`, set when the
buffer starts filling and cleared when it empties, so `stream_dispatch_staged`
takes only the engine and passing the wrong array is no longer expressible.
`dispatched_elements()` is gone: it subtracted an engine-wide byte count from a
per-array cursor, which meant nothing unless the precondition held. Empty
staging returns `writer_ok()`, and the `append_failed` test moves to
`stream_flush_body`, the only caller that reached the dispatch without one.

Refactor only, no behaviour change.

### DIFF
--- a/src/gpu/compress.cu
+++ b/src/gpu/compress.cu
@@ -99,7 +99,7 @@ codec_alignment(enum compression_codec type)
 extern "C" size_t
 codec_output_alignment(enum compression_codec type)
 {
-  nvcompAlignmentRequirements_t written = { 0 }, read = { 0 };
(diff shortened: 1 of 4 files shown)


nclack:
The dispatch also reads the engine's bound per-array state, so staging must
empty before another array binds. Naming the owner does not change that, and
both multiarray pre-dispatch calls stay.

PR #186, opened 2026-08-08
One scatter launch per dispatch

### LINKED ISSUE #181: One scatter launch per dispatch instead of one per epoch
A dispatch can cover several epochs, and the scatter handles that by issuing one
kernel launch per epoch, walking a device pointer forward one epoch region at a
time. The kernel already has what it needs to do the whole buffer in one launch.

`transpose_v0_k` computes each element's destination from its absolute index,
decomposed against the lifted shape. That decomposition repeats every epoch, which
is exactly why the caller has to hand it a fresh base pointer per epoch. Give the
kernel the epoch size and the distance between epoch regions and it can add
`(index / epoch_elements) * epoch_stride` itself, so one launch covers the buffer
whichever epochs it spans. `struct scatter_destination` already carries both
numbers.

The reason to consider it is not speed. Scatter runs at 77 GiB/s on 256cube and
91 GiB/s on smallepoch, and the many-launch geometry is the faster of the two, so
there is no evidence the launches cost anything. It is that the slicing is what
forces everything else:

- `scatter_by_epoch` and its pointer walk, which is a stage stepping through pool
  memory by hand rather than through the pool API
- an unaligned source, since a slice starts partway into the staging buffer
- the kernel's word-aligned load path and the lead/surplus masking it needs
- `TRANSPOSE_SOURCE_PAD_BYTES`, its allocation slack, and the zeroing at init
- the transpose tests that cover every sub-word start offset

All of that exists to support slicing. One launch removes it.

The measurement to make first is whether a single launch is slower, since it adds
one division per thread to a kernel that already does about forty. If it is within
noise, the simplification is free.

nclack:
A dispatch can cover several epochs. The scatter issued one kernel launch per
epoch, walking a device pointer forward one region at a time, and that slicing
forced everything around it: a stage stepping through pool memory by hand, a
source starting partway into the staging buffer, the kernel's word-aligned load
path with its lead handling, and the padding and zeroing that kept its reads in
bounds.

`transpose_v0_k` already computed each element's destination from its own index.
It now also takes the epoch size and the distance between epoch regions, so it
adds the region step itself and one launch covers the buffer whichever epochs it
spans. The check against the regions the caller holds moves to the host, where
it is one comparison instead of one per epoch and runs before the transfer is
queued.

With the source always the staging base, the kernel reads it directly. That
removed the shared-memory staging, the word load with its lead and surplus
handling, `TRANSPOSE_SOURCE_PAD_BYTES`, the allocation slack, and the memset at
init. The transpose tests that covered every sub-word start offset now cover
several epochs in one call, for each element size, starting both on an epoch
boundary and one element in. Destination regions in the tests are padded, so a
kernel that stepped by epoch instead of by the stride it is handed no longer
passes.

One launch is faster, not slower: the launches cost more than the per-element
division that replaces them. On one L40 the scatter stage goes 87 to 188 GiB/s
on smallepoch and 46 to 56 GiB/s on 256cube.

Closes #181

### DIFF
--- a/dev/gpu-orchestration.md
+++ b/dev/gpu-orchestration.md
@@ -79,11 +79,11 @@ data itself.
 **Dispatch groups appends** (#173). An append copies into the pinned staging
 buffer and returns. Nothing reaches the device until that buffer fills or the
 caller flushes, at which point one transfer moves the whole buffer and one
-scatter runs per epoch it covers. A dispatch is capped at the room left in the
-batch, since that is what the chunk pool holds — or at one epoch for a
-multiscale stream, whose linear buffer holds one. So `buffer_capacity_bytes`
-sets the transfer size, and the append cursor runs ahead of the epochs the
-schedule has counted by whatever is still staged.
+scatter covers it, whichever epochs it spans. A dispatch is capped at the room
+left in the batch, since that is what the chunk pool holds — or at one epoch
+for a multiscale stream, whose linear buffer holds one. So
+`buffer_capacity_bytes` sets the transfer size, and the append cursor runs
+ahead of the epochs the schedule has counted by whatever is still staged.
 
 ### Cost of grouped dispatch
 
--- a/dev/gpu-orchestration.md
+++ b/dev/gpu-orchestration.md
@@ -284,12 +284,13 @@ pointer with no ordering attached, and has seven non-test callers in
 Each is correct today because a comment says so. Two zero a pool; the rest read
 at an offset inside a generation the caller already holds.
 
-#173 added a second way past the pool API — the scatter stepped a raw device
+#173 added a third way past the pool API — the scatter stepped a raw device
 pointer from one epoch's region to the next — and #181 removed it by scattering
 a whole buffer in one launch. The kernel derives the region from the element
-index and the stride it is handed, so nothing steps pool memory by hand.
+index and the stride it is handed, so nothing steps pool memory by hand. The two
+`gpu_pool_at` uses above remain.
 
-Work: give the pool an operation for the remaining use, then remove
+Work: give the pool an operation for each of those two uses, then remove
 `gpu_pool_at`.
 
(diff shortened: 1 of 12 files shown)


nclack:
Measured on one L40, RelWithDebInfo, scatter stage, two reps each, before and
after in the same run.

| bench | dtype | main avg GiB/s | one launch | main best | one launch |
| --- | --- | --- | --- | --- | --- |
| smallepoch, 200k frames | u16 | 86.6, 87.5 | 188.0, 188.5 | — | — |
| smallepoch, 200k frames | u8 | 87.5, 87.5 | 189.3, 189.0 | — | — |
| 256cube, 2k frames | u16 | 45.9, 45.7 | 55.1, 55.7 | 146.2, 145.8 | 141.8, 142.6 |
| 256cube, 2k frames | u8 | 45.8, 45.6 | 55.1, 55.7 | 145.9, 145.4 | 141.9, 142.4 |

A dispatch on smallepoch spans many epochs, so it paid many launches. The added
division shows up only in the best single launch, 2.6% slower. smallepoch records
one sample per run, so its best equals its average.

One suggested cleanup was a 12% regression. The 4KB block constant outlived the
shared buffer it sized and was duplicated on host and device, so the obvious fix
was to drop the tiling for one element per thread. Measured, that costs the whole
gain: 256cube fell to 49.4 GiB/s, level with main. The block loop gives the
compiler several destination computations to overlap. The tiling stays, and both
sides now read one `constexpr`.

Nothing gave the kernel a region stride that differed from an epoch's elements.
`build_lifted_layout` sets the chunk stride to the chunk's element count, while
production pads chunks to the codec's alignment, so a kernel stepping regions by
epoch would have passed the whole suite. The tests pad the region now.

The tests' `first_element` generality was dead. The kernel only ever sees the
offset within the epoch, so a call site past the first epoch exercises nothing
new.

PR #187, opened 2026-08-09
Build the benchmarks on machines without a GPU

### LINKED ISSUE #97: build benchmarks on non-gpu platforms

nclack:
The benchmarks only built when CHUCKY_ENABLE_GPU was on, so a CPU-only
checkout had none. bench/ is now added unconditionally.

The GPU pipeline sits behind bench/bench_gpu.h, which has no CUDA in it.
bench_gpu.cuda.c forwards to tile_stream_gpu; bench_gpu.none.c stands in
when GPU support is off, and bench_util.c no longer reaches for cuda.h
or stream.gpu.h. The two-stream benchmark needs a GPU, so it is built
only where there is one, the way the GPU-only tests already are.
Without GPU support the --backend default is cpu, and asking for
--backend gpu fails with a message to reconfigure.

The batch-size target in the stream configuration is now 64 bits rather
than 32, so a batch above 4 GiB works and nothing has to be narrowed on
the way in. The benchmark command line reads sizes with fixed-width
types throughout, and --chunk-bytes, --batch-bytes, --memory-budget and
--backpressure now reject text they cannot read instead of parsing it as
something else. --batch-bytes also reaches runs that skip the auto-fit
path, which ignored it.

Closes #97.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -33,9 +33,7 @@ add_subdirectory(src)
 enable_testing()
 add_subdirectory(tests)
 
-if(CHUCKY_ENABLE_GPU)
-    add_subdirectory(bench)
-endif()
+add_subdirectory(bench)
 
 # --- Install rules: aggregate library, public headers, CMake/pkg-config ---
 
--- a/bench/CMakeLists.txt
+++ b/bench/CMakeLists.txt
@@ -45,8 +45,7 @@ target_link_libraries(
     PRIVATE warnings
 )
 
-# The GPU pipeline sits behind bench_gpu.h so the rest of the benchmarks build
-# without CUDA. bench_gpu.none.c reports "built without GPU support" instead.
+# Without CUDA the benchmarks link a stand-in that fails at startup.
 if(CHUCKY_ENABLE_GPU)
     target_sources(bench_util PRIVATE bench_gpu.cuda.c)
     target_link_libraries(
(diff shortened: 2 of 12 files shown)


nclack:
Verified on the cluster, both configurations built from the same tree.

CPU-only (`-DCHUCKY_ENABLE_GPU=OFF`, conda deps, no CUDA on the path):
all 12 bench executables link warning-free, ctest passes 31/31,
`bench_stream_256cube_single --frames 8` runs end to end at 3.55 GiB/s,
and `--backend gpu` and `bench_stream_256cube_two_streams` both exit
non-zero with the reconfigure message.

GPU (L40, sm_89, CUDA 13.1): `bench_stream_256cube_single`,
`bench_stream_smallepoch_single`, the same binary with `--backend cpu`,
and `bench_stream_256cube_two_streams` all pass. The memory breakdown and
per-stage tables print the same lines as before, since the reporting moved
into bench_gpu.cuda.c unchanged.

The two nvcomp `-Wmissing-field-initializers` warnings in
src/gpu/compress.cu predate this branch.

An alternative was to skip the GPU-only benchmarks in a CPU-only build, the
way tests/CMakeLists.txt skips GPU tests. Building all of them and failing
at startup keeps one target list on every platform, and CI's existing
CPU builds on Linux, macOS, and Windows then cover the whole bench
directory.

PR #188, opened 2026-08-09
Show readers only finalized data

### LINKED ISSUE #123: Streaming readers can see invalid state: unfinalized shards advertised + torn zarr.json
## Goal

A reader/visualization tool opening the zarr **mid-stream** should always see a valid view: every byte advertised by metadata is in a finalized, parseable shard, and metadata itself is never observed mid-write.

Today, two gaps break this contract.

## Gap 1 — metadata advertises chunks in unfinalized shards

The periodic `update_append` callback writes `zarr.json` with append-dim sizes derived from:

```
flat_append_chunks = shard_epoch * chunks_per_shard_append + epoch_in_shard
```

- GPU path: `src/gpu/flush.d2h_deliver.c:350-379`
- CPU path: `src/cpu/stream.body.c:201-218`

`epoch_in_shard` counts chunks already delivered into the **currently open** shard, which has not been finalized — its index block isn't on disk until `finalize_shards` runs (`src/zarr/shard_delivery.c:35-102`).

Concretely, a reader that:
1. Reads the latest `zarr.json` and computes the implied chunk grid,
2. Opens the corresponding shard file,
3. Locates the index at \`file_size - index_total_bytes\`,

will land **inside the data region**, not on a valid index, for any shard not yet finalized. Result: garbage chunk offsets/sizes, corrupted reads.

This is also the open question raised in #122.

## Gap 2 — \`zarr.json\` writes are non-atomic

`fs_put` (`src/zarr/store_fs.c:24-41`) opens the metadata file with \`O_WRONLY | O_CREAT | O_TRUNC\`, writes, then closes. A concurrent reader can observe:

- Zero-length file (between \`O_TRUNC\` and the first write completing)
- Partial JSON (mid-write)
- Stale JSON (if the writer fails mid-write and leaves a truncated file)

Any of these break a streaming reader's parse step before they even get to shard files.

### Fix 1: clamp \`update_append\` to the last completed shard boundary

Advertise only sizes derived from \`shard_epoch * chunks_per_shard_append\` — i.e., drop the \`epoch_in_shard\` term. Metadata then only ever describes data that is already finalized on disk.

Cost: append-dim shape lags the actual write cursor by up to one shard's worth of chunks. For typical \`chunks_per_shard_append\` values this is acceptable; the reader sees stair-step growth instead of smooth growth.

If smoother growth matters, an alternative is to also flush the in-progress shard's index (and re-open / re-extend) on each metadata update — but that defeats the streaming-write design and isn't recommended.

### Fix 2: atomic-rename \`zarr.json\` writes

In \`fs_put\` (and the equivalent S3 path if applicable), write to \`<key>.tmp.<unique>\`, \`fsync\`, then \`rename\` over the final path. POSIX guarantees \`rename\` is atomic within the same filesystem. Win32: \`MoveFileEx(MOVEFILE_REPLACE_EXISTING | MOVEFILE_WRITE_THROUGH)\`.

This is a small, contained change to one function plus a platform helper.

## Out of scope

- Shard file size tightness — see #122.
- Recovery / restart of streams — separate concern; relies on shard finalization being durable.

## Test gaps

- Add a stress test that reads \`zarr.json\` continuously while a stream writes, asserting it always parses.
- Add a test that opens an in-progress shard at the advertised size and verifies the index is parseable (gates the Fix 1 invariant).

### LINKED ISSUE #175: Shape is not caught up when a flush fails
After a flush fails, the array's recorded shape stays at whatever the last
periodic metadata update wrote. Shards that were delivered and finalized after
that point are on disk but invisible to a reader, because nothing has extended
the shape to cover them.

The final flush is the only thing that catches the shape up to the end of the
stream — `update_append`'s contract says it is called periodically during
streaming and at final flush — and on the failing path it is skipped. With
`metadata_update_interval_s` at its default the gap can be a whole interval's
worth of shards.

The shape is skipped rather than written because it is derived from the append
cursor, which counts everything the caller handed over, including data still in
staging and batches that failed to deliver. Writing it after a failure would
claim frames that are not there, which is worse than leaving it short.

This is not specific to a failed flush: the same derivation means the shape a
successful flush writes is only right because everything did arrive.

nclack:
A reader that opens the zarr while a stream is running could see two
kinds of invalid state, and a failed flush could hide data that was
already on disk.

The periodic metadata update derived the append extent from the write
cursor, so it counted chunks in a shard that had not been closed out.
That shard has no index block yet, so a reader following the advertised
extent into the file landed in the data region and read garbage offsets.
The extent now comes from the append chunks in closed-out shards whose
writes have retired, so it grows a shard at a time and lags the write
cursor.

fs_put opened zarr.json with O_TRUNC and wrote in place, so a reader
could see an empty file or a partial document. It now writes a sibling
temporary file and renames it over the target. The rename is what makes
the swap atomic; the bytes are not flushed to the device first, because
shard data is not flushed either. S3 needs no change, since a PutObject
replaces the object in one step.

The final flush skipped the shape whenever anything upstream failed,
leaving shards finalized since the last periodic update on disk but
invisible. The shape now names only finalized shards, so it is written
on the failing path too, held down to the elements actually appended so
a padded final chunk is not reported as data. A sink IO error stays the
exception, because which writes landed is then unknowable.

Draining the kicked slots continues past the first failure. A slot left
kicked keeps a delivery worker running that owns the shard state the
flush goes on to read.

The benchmark metering sink forwards update_append, so an -o run
exercises the metadata writes it used to skip.

Closes #123. Closes #175.


nclack:
Both issues have the same root cause, so they are fixed together: the shape came from the append cursor, which counts everything the caller handed over. Deriving it from the delivery counters instead fixes the mid-stream over-claim in #123 and removes the reason the final flush treated the shape as a claim that had to be earned in #175.

The count is held to the elements actually appended. Shard counters alone are chunk-granular, so a stream that ends mid-chunk would report the padding as data — a 5-frame stream with a 4-frame chunk would say 8. Taking the smaller of the two keeps the shape on a clean stream byte-identical to before.

The count only advances once the finalize's writes have retired, using a fence recorded at finalize time. The pipeline has normally retired it long before the next metadata update asks, so the wait costs nothing in the steady state.

Verification on an L40 (sm_89, RelWithDebInfo): 51/51 ctest pass, no new compiler warnings.

Reverting only `src/` and keeping the new tests, all four affected tests fail, each for the intended reason:

- `test_store_fs` — 172 of 395 concurrent reads torn
- `test_zarr_fs_sink` — mid-stream shape `[8,8,12]`, naming a shard with no index, instead of `[6,8,12]`
- `test_shape_after_flush_error` — no shape written at all after a failed flush
- `test_batch` — same

`test_batch`'s `failed_delivery_writes_no_shape` asserted the old contract. Its reason was that the shape would claim frames that never reached the sink, which no longer applies, so it now asserts the shape is written and names nothing.

nclack:
Windows needs a retry around the replace. MoveFileEx cannot replace a file another handle has open unless that handle allowed delete sharing, which readers do not, so `platform_rename_replace` waits out `ERROR_SHARING_VIOLATION` for up to 100 ms. The concurrent-reader test polls instead of holding the file open continuously, which no real reader does; without the fix it still sees 101 of 101 reads torn.

All six checks pass.

nclack:
## Filesystem output benchmark

L40, RelWithDebInfo, output to node-local xfs (`-o /tmp/...`), merge base 020f219 vs this branch. Each A/B ran both binaries interleaved in one allocation from one build directory, so binary layout and machine drift hit both.

| config | merge base | this branch | delta |
|---|---|---|---|
| 256cube, 200 frames | 8.960 GiB/s | 8.918 GiB/s | -0.46% |
| smallepoch, 2M frames | 3.873 GiB/s | 3.744 GiB/s | -3.33% |

Medians of 9 reps. Run-to-run spread on smallepoch is ±0.06–0.14 GiB/s, so treat -3.33% as approximate.

The data path is untouched, and every GPU stage (memcpy, h2d, scatter, compress, aggregate, d2h, sink) matches within noise on both configs. The difference sits in producer flush stall.

### Metadata write volume goes down

The advertised extent now changes only at shard boundaries, and `zarr_array_update_append` skips the rewrite when the extent is unchanged, so the branch writes `zarr.json` less often. Counted with `strace -f -e trace=openat,rename`:

| config | merge base | this branch |
|---|---|---|
| 256cube | 27 rewrites | 8 rewrites |
| smallepoch | 5 rewrites | 5 rewrites |

Cost of one rewrite, measured on an idle filesystem: 114 µs truncate-and-write versus 235 µs write-temp-and-rename on xfs, and 2.76 ms versus 5.53 ms on the shared NFS home. Writing to NFS is roughly 25x the cost of node-local xfs either way.

### Why there is no fsync

An earlier revision flushed the temporary file before renaming. That is dropped. Shard data is never flushed to the device, so flushing only the metadata would buy nothing except a journal commit mid-stream. Reader atomicity comes from the rename, not the flush.

### Remaining question

The smallepoch cost is about 10 ms on a 250 ms run. Two candidate causes were tested and neither accounts for it: making the fence check non-blocking changed nothing (-4.57% versus -3.99% in the same allocation), and removing the fsync recovered at most a third. smallepoch is the extreme end of the range — one chunk per epoch, about 5 batches and 7 shard files for the whole run — so a few milliseconds anywhere shows up as several percent. 256cube, with 25 batches and 82 shard files, is flat.

### Benchmark harness

`metering_sink_init` did not forward `update_append`, and every `-o` and `--s3-*` run wraps the real sink in it. With it NULL the stream skips both the periodic metadata update and the final shape publish, so a `-o` bench wrote `zarr.json` only at array creation — two writes for a whole run. None of the numbers above were measurable before that was fixed. The fix is in this branch (`Meter sink forwards update_append`) and it changes what every `-o` benchmark measures, so sweeps from before it are not comparable for anything metadata-related.

PR #189, opened 2026-08-09
Lower staging copy threshold to 64 KiB

### LINKED ISSUE #183: Staging copy threshold no longer matches append sizes
`ingest_copy` hands a staging copy to the thread pool only when it is at least
2 MiB. Since the writer started grouping appends (#173) that threshold no longer
matches the traffic: an append is one frame, so 128 KiB on the 256cube benchmark
and 512 bytes on smallepoch, and the pool is never used on that path. It is used
for the large-block path, which is not the shape the writer is for.

Measured on one L40, three interleaved repetitions, clocks pinned. The copy is
the same bytes in every row; only the path differs.

| benchmark | append size | copy path | memcpy stage |
|---|---|---|---|
| 256cube | 128 KiB | single thread | 1331 ms, 14.1 GiB/s |
| 256cube | 64 MiB | thread pool | 410 ms, 46.2 GiB/s |
| smallepoch | 512 B | single thread | 47.6 ms, 20.0 GiB/s |
| smallepoch | 64 MiB | thread pool | 17.7 ms, 54.0 GiB/s |

The pool is three times faster per byte. Its cost is a mutex, a broadcast and
some atomics per dispatch, on the order of 100-300 ns against a 128 KiB copy that
takes about 9 microseconds, so a threshold somewhere near 64-128 KiB looks
defensible. 512 bytes does not, so keep a floor.

Two things to weigh in the measurement. The helpers compete with compression for
cores and cache, and on 256cube the compressor is the pace setter at about 1.6
seconds, so a producer-side win there buys headroom rather than wall time. And
the copy destination is now swept rather than reused, which is a separate cost:
appends used to land in the same first bytes of the pinned buffer and stay in
cache, which is why this stage reads 584 ms on the commit before #173.

Write-combined staging memory was measured as an alternative and rejected. It
recovers a good part of the single-threaded loss (1331 to 812 ms on the 128 KiB
case) but costs the pooled path 78% on smallepoch and 6% of end-to-end
throughput, and the flag is fixed at allocation so one buffer cannot serve both.
Full numbers are in the pull request discussion for #173.

nclack:
ingest_copy handed a staging copy to the thread pool only at 2 MiB and up.
An append has been one frame since #173, so the writer's own traffic never
reached that size and every staging copy ran on the producer thread.

The threshold is now 64 KiB. A dispatch costs about 1.4 us, so the split
starts to pay between 16 and 32 KiB, and 64 KiB keeps margin. On one L40 the
staging copy runs about twice as fast at frame-sized appends. Appends below
the threshold and above 2 MiB are unchanged.

Also clears the build warnings: the nvcomp alignment structs are
value-initialized rather than partly initialized, and the threadpool test
casts away _Atomic before free, which MSVC reported as C4090.

Adds a compress-tier sweep for reef-l40, whose last entry was five weeks old.

Closes #183


nclack:
Measured on one L40, 16 cores, `--max-threads 16`, three interleaved
repetitions, median. The memcpy stage is the producer's staging copy.

| benchmark | frames | append | main | this change | main GiB/s | this change |
| --- | --- | --- | --- | --- | --- | --- |
| smallepoch | 2M | 512 B | 47.5 ms | 47.4 ms | 20.1 | 20.1 |
| 256cube | 500 | 32 KiB | 1626 ms | 1618 ms | 14.4 | 14.5 |
| 256cube | 500 | 64 KiB | 3372 ms | 1731 ms | 13.9 | 27.1 |
| 256cube | 500 | 128 KiB | 3387 ms | 1372 ms | 13.8 | 34.2 |
| 256cube | 2000 | 128 KiB | 13670 ms | 5496 ms | 13.7 | 34.1 |
| smallepoch | 2M | 61 MiB | 16.6 ms | 15.9 ms | 57.4 | 60.0 |
| 256cube | 2000 | 64 MiB | 3173 ms | 3415 ms | 59.1 | 54.9 |

64 KiB is the first size the change reaches. Rows below 64 KiB and above 2 MiB
run the same code in both builds, so their spread sets the noise floor: the
64 MiB repetitions run 3.1-4.1 s either way, making the 7% gap in that row
noise, while the 64 KiB and 128 KiB repetitions hold within 0.5%.

End-to-end throughput moves only where the copy is the slowest stage.

| append | main | this change | slowest stage |
| --- | --- | --- | --- |
| 512 B | 3.00 GiB/s | 3.00 GiB/s | compression |
| 32 KiB | 11.45 GiB/s | 11.48 GiB/s | compression |
| 64 KiB | 12.35 GiB/s | 20.76 GiB/s | the copy |
| 128 KiB | 11.03 GiB/s | 11.15 GiB/s | compression |

At 64 KiB, halving the copy cut wall time from 3.80 to 2.26 s and median append
latency from 5.0 to 2.2 µs.

One caveat: `pump_data_prefill_blocked` fills a single append-sized block and
resends it, so the data repeats with a period equal to the append size and
larger appends compress worse — 155x at 32 KiB, 120x at 64 KiB, 31x at 128 KiB.
That is why compression is the slowest stage in some rows and not others. Each
row compares the same data against itself, so read down a column, not across
rows.

nclack:
Why 64 KiB. A build that always splits the copy against one that never does,
same L40 and 16 cores, two repetitions, median of the memcpy stage. The pool
splits four ways, so `dispatch = pooled - serial/4`.

| append | serial | pooled | ratio | dispatch |
| --- | --- | --- | --- | --- |
| 512 B | 20.1 GiB/s | 0.6 GiB/s | 0.03 | 0.74 µs |
| 2 KiB | 61.0 GiB/s | 2.0 GiB/s | 0.03 | 0.95 µs |
| 8 KiB | 33.9 GiB/s | 6.1 GiB/s | 0.18 | 1.19 µs |
| 16 KiB | 16.6 GiB/s | 12.1 GiB/s | 0.73 | 1.03 µs |
| 32 KiB | 14.8 GiB/s | 20.0 GiB/s | 1.35 | 1.01 µs |
| 64 KiB | 14.2 GiB/s | 27.5 GiB/s | 1.94 | 1.14 µs |
| 128 KiB | 14.0 GiB/s | 34.2 GiB/s | 2.45 | 1.38 µs |

The curves cross between 16 and 32 KiB. Dispatch costs 1.0-1.4 µs at every
size, and that is what sets the crossover; the issue's estimate of 100-300 ns
was low by about 5x. The code comment now carries the measured figure.

64 KiB sits about 2.5x above this crossover, which is less margin than it
sounds: the crossover moves with the producer's rhythm more than with the
machine, and rises past 256 KiB once the helpers sleep. What picks 64 KiB over
the 32 KiB that also wins here is the pool size — with a single helper a 32 KiB
split loses and a 64 KiB one wins — and the ceiling at the other end, since
anything above 128 KiB would miss the frame-sized appends this is for. Below
8 KiB the pool is not a small loss but a large one, 30x at 512 B, so the floor
has to be there.

The threshold is a minimum, so raising it pools less: 512 KiB would still leave
the 128 KiB frame-sized appends on the producer, which is the bug being fixed.

nclack:
The sweep in `bench/results/reef-l40-924bdb0-20260809.json` passes 96 of 96
runs. It is a refreshed baseline, not a before-and-after for this change: no
sweep run passes `--append-elements`, so every one appends the default 64 MiB
block, above both the old threshold and the new one.

The last reef-l40 entry was 46fa2e8 from five weeks ago, so comparing against it
would fold in everything merged since. The no-regression evidence is the matched
A/B instead: at 64 MiB and 61 MiB appends, where both builds run the same code,
the copy stage sits still.

The sweep matrix has no append-size dimension, which is the same blind spot that
let this bug sit. Every run measures 64 MiB blocks while the writer sends one
frame per append.

nclack:
This branch clears the build warnings. A from-scratch GPU build and a
from-scratch CPU-only build each report zero, and MSVC no longer reports C4090
on the runner that reported it. The macOS, cuda12, cuda13 and cpu-tests jobs
were already clean.

The numbers above were measured against 020f219. #186 and #184 have landed
since, and #186 makes the scatter faster, so which stage is slowest may have
shifted and the end-to-end column could read differently now. Neither touches
ingest_copy or the producer's copy, so the memcpy comparison stands. This branch
still merges into main without conflict.

nclack:
Where the 15 µs in the code comment comes from.

The crossover above was measured with appends back to back, which keeps the
pool's helpers spinning. Let them go to sleep and the dispatch costs more. Same
copy, varying only the idle gap before it, median µs per copy, three helpers
plus the producer:

| payload | no gap | 50 µs gap | 500 µs gap | 5 ms gap |
| --- | --- | --- | --- | --- |
| 32 KiB | 0.65 | 1.06 | 0.09 | 0.08 |
| 64 KiB | 1.33 | 1.62 | 0.15 | 0.12 |
| 128 KiB | 2.09 | 2.52 | 0.26 | 0.19 |
| 1 MiB | 7.68 | 14.21 | 6.47 | 2.15 |

Each cell is serial time over pooled time, so above 1 the pool wins. Once the
helpers sleep, the pooled path costs a flat 4 µs at a 500 µs gap and 15 µs at
5 ms whatever the size, which is waking them and nothing else. That moves the
size where splitting pays from about 24 KiB up past 256 KiB.

So 64 KiB is right for a producer that appends steadily and wrong for one that
appends 64 KiB to 512 KiB with more than about 500 µs between calls. In that
band the split costs roughly 10 µs against a gap of at least 500 µs, under 2% of
the interval that created it, so it cannot hold back a frame rate. Raising the
threshold to clear the sleeping case would put it above the frame-sized appends
this change is for.

Large payloads are clear either way. At 1 MiB the pool wins 2.15x even after
5 ms of idle, because the copy dwarfs the cost of waking them — and 1 MiB is
below the old 2 MiB threshold, so that case was copying on the producer before
this change.

PR #191, opened 2026-08-10
Fewer divisions in the scatter kernel

### LINKED ISSUE #190: Scatter kernel divides more than it needs to
`transpose_v0_k` spends most of its per-element work on integer division. It
derives each element's destination by decomposing that element's index against
the lifted shape, which costs one 64-bit modulo and one division per lifted
dimension. The lifted rank is twice the array rank, so a 3D or 4D array pays 12
to 16 divisions per element, plus one more for the region step added by #181. The
kernel also reads `shape` and `strides` from global memory inside that loop, once
per dimension per element.

Compiled for sm_89 with nvcc 13.1 at -O3, the u16 instantiation is 584
instructions and holds 40 registers. Fourteen loads in the per-element loop are
`LDG.E.64`, about 128 bytes of memory traffic to place one 2-byte element, and
eleven of the divisions call the 64-bit division subroutine instead of inlining.

This is not about stream throughput. The scatter stage is far from the bottleneck
and none of it will move an end-to-end benchmark. The register count is the part
worth fixing: the kernel's only job is an indexed copy, and it holds 40.

nclack:
The scatter kernel spent most of its per-element work on integer division,
and read the shape and strides from global memory once per dimension per
element.

Three changes. The host now finds the run of trailing extents that makes up
one epoch, so dividing an element index by them leaves the epoch number.
That drops the separate division by epoch_elements along with the iterations
for the dimensions in front, whose coordinates cannot reach the destination.
The shape and strides travel by value in the launch parameters instead of
through device copies, which retires struct tile_stream_layout_gpu and its
per-level allocations. A 32-bit index instantiation runs whenever every index
the kernel forms provably fits, which inlines the per-element divisions.

The scatter now returns a status, because it can refuse a layout it cannot
place and it reports a launch the driver turns away.

test_transpose and test_ingest staged their input with legacy-stream calls
while launching on non-blocking streams. The shape and stride uploads were
the only thing ordering the two, so removing them exposed the race and the
kernel read its source before the copy landed. Both tests now stage on the
stream their kernel runs on. New cases cover two append dimensions, the wide
index kernel, and the layouts the scatter refuses.

Closes #190

### DIFF
--- a/src/gpu/stream.c
+++ b/src/gpu/stream.c
@@ -96,7 +96,6 @@ engine_dispatch_ingest(struct stream_engine* e,
     };
     return ingest_dispatch_scatter(&e->stage,
                                    &ctx->layout,
-                                   &ctx->layout_gpu,
                                    dst,
                                    first_element,
                                    dtype_bpe(ctx->config.dtype),
--- a/src/gpu/stream.engine.h
+++ b/src/gpu/stream.engine.h
@@ -121,7 +121,6 @@ struct lod_state
 
   // Per-level chunk layouts [0..nlod-1]
   struct tile_stream_layout layouts[LOD_MAX_LEVELS];
-  struct tile_stream_layout_gpu layout_gpu[LOD_MAX_LEVELS];
 
   // Morton-to-chunk scatter LUTs (precomputed)
   CUdeviceptr d_morton_chunk_lut[LOD_MAX_LEVELS];
@@ -280,7 +279,6 @@ struct stream_context
   struct tile_stream_configuration config;
   struct shard_sink* sink;
   struct tile_stream_layout layout;
-  struct tile_stream_layout_gpu layout_gpu;
   struct level_geometry levels;
   struct dim_info dims;
   uint64_t cursor_elements;
--- a/src/gpu/stream.ingest.c
+++ b/src/gpu/stream.ingest.c
@@ -122,8 +122,42 @@ scatter_timing_begin(struct staging_state* stage, size_t bytes)
     stage->scatter_samples_lost++;
   stage->next_timing = (stage->next_timing + 1) % SCATTER_TIMING_SLOTS;
   st->bytes = bytes;
-  st->pending = 1;
+  st->pending = 0;
   return st;
+}
+
+// The measurement only counts once both ends are recorded: an interval with one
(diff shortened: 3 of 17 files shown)


nclack:
Measured on an L40 (sm_89), RelWithDebInfo, u16, two reps, against a build of
main at 1eb2906.

| u16 kernel | main | here |
| --- | --- | --- |
| instructions | 584 | 248 narrow, 520 wide |
| registers | 40 | 28 narrow, 36 wide |

| scatter stage, GiB/s (avg / best) | main | here |
| --- | --- | --- |
| `bench_stream_256cube_single`, 2k frames | 55.8 / 142 | 96.5 / 232 |
| `bench_stream_smallepoch_single`, 200k frames | 187 / 187 | 265 / 265 |

End to end is flat, which is what #190 predicted: 256cube reads 9.7 GiB/s
before and 9.86 after. Use those frame counts to see the stage at all — the
400-frame default records a single sample.

The split is derived per dispatch rather than stored beside the layout, so a
layout the scatter cannot place is refused at a dispatch instead of at open.
Deriving it costs about 2·lifted_rank host divisions against a launch, so the
argument for moving it is where the failure surfaces, not speed. Doing it
properly means a field on `struct tile_stream_layout` filled in
`compute_level_layout`, which would also make both refusals unreachable and
let the scatter take the split instead of `epoch_elements`. Worth its own
change.

Three more things this leaves behind. `ingest_dispatch_multiscale` still fails
between its two timing events while holding the pool slot, which is the leak
this PR fixed for the scatter path. This is the only one of 20 kernel launches
in `src/gpu` that reads the runtime error, so it has to clear the slot first to
avoid reporting a neighbour's; a shared launch check at every site would retire
the clear. And `aggregate_layout` still uploads and charges for device copies of
its lifted shape and strides that no kernel reads — the same thing this PR
removed, one file over.

Coverage that is still missing: a destination offset past UINT32_MAX picks the
wide kernel, and reaching it needs a 4 GiB destination, so no test does.

Verified before merge: 50/50 ctest on an L40 with a warning-free build.

nclack:
The follow-ups listed above are now #195, along with one more: the test layout helper builds an unpadded chunk stride, so no GPU kernel test scatters with production geometry.

PR #194, opened 2026-08-10
Flush finalizes the stream

### LINKED ISSUE #193: Append extent stops short of data written after a mid-stream flush
A flush that lands partway through a chunk pads that chunk out and closes the shard. Anything appended afterwards starts past the padding, and past the shard slots the flush left empty. The published append extent stops short of those chunks, so frames that are on disk with a valid shard index cannot be read.

The extent is the smaller of two numbers: how far along the append dim the finalized chunks reach, and how many elements the caller appended. After a mid-stream flush the element count is the smaller of the two, and it no longer describes where the data ends, because the padding and the empty slots sit between the appended elements and the grid position they occupy.

## Reproducing

CPU stream, append dim `chunk_size = 2`, `chunks_per_shard = 2`, one chunk per append:

1. Append one chunk, then `writer_flush`. Shard 0 closes holding chunk 0, its second slot unused.
2. Append two more chunks, then flush.

Those two chunks occupy append positions 4 through 7. The published shape is 6, so the last chunk is outside the array bounds.

The padded-chunk variant behaves the same way: with `chunk_size = 3`, appending one plane, flushing, then appending one more plane leaves real data at position 6 and publishes a shape of 2.

## Scope

This predates #188. Before it the final shape came from the append cursor alone, which hides the same frames; #188 additionally bounded the extent by the finalized chunks, which does not change this case. Both backends are affected, and every level of a multiscale pyramid.

`writer_flush` is documented as leaving the stream appendable and is covered by `test_append_after_flush`, so appending after a flush is a supported sequence.

## Test coverage

No test covers the shape published when appends follow a flush, on either backend. Tests written for it during #188 asserted values that hid data, so they were removed rather than merged.

nclack:
A flush pads the chunk the append cursor stopped in and closes its shard, which
is what makes that data readable. Anything appended afterwards had to start past
the padding and past the shard slots the close left empty, so it landed at append
positions the caller never asked for, with the pre-flush frame written a second
time. The published extent stopped short of those chunks, so frames sat on disk
with a valid shard index and no reader could reach them.

Flush now finalizes the stream. It writes out everything appended, including that
partial chunk, and then refuses input: append consumes nothing and reports
finished. It is idempotent, and a flush that fails partway finalizes too, since
it may already have closed shards. Later calls re-report the same outcome.

That removes the case rather than computing around it. With no appends after a
flush there is only ever one padded chunk, the last, so the extent stays the
smaller of the finalized chunk reach and the elements appended, unchanged from
#188. No padding bookkeeping is needed, and none is added.

Flush also stops waiting for those writes to land. It returns once they are
queued; the new close call waits for them, publishes the extent and runs the
sink's own flush, and reports whether all of it succeeded. So close is where a
caller learns its data reached storage and where the array becomes readable, and
the sink has to outlive the stream: destroy runs close only when the caller did
not. The gain is that the multiarray frontends queue every array before waiting,
so their writes overlap instead of draining one array at a time.

The contract is stated on the writer vtable, in the design doc and in the guide.
Tests follow it: the two that asserted a resumable flush now assert a finalizing
one, the three that asserted completion moved to close, and every clean case in
the shape tests checks that the append after the flush is refused.

Closes #193.


nclack:
## Why this one and not the four in #193

All four earlier attempts tried to work out where the data ends from counters
taken after the fact, and none of those counters sees the padding. This records
the padding at the moment it is created, so nothing has to be reconstructed.

- Counting chunks, or counting skipped slots, still misses padding *inside* a
  chunk. Here the padding is the recorded quantity, and the grid reach is
  already correct on its own.
- The "a shard closed early" flag became the default path because the ordinary
  end-of-stream flush also closes a partial shard. This is a number, not a flag:
  a flush that pads nothing records zero and changes nothing.
- The `appended_after_flush` latch was set in the writer wrappers, which the
  multiarray frontends bypass. The record happens in `cpu_stream_flush_body` and
  `stream_flush_body` instead, and every frontend — single stream, multiarray,
  destroy-time auto-flush — goes through those.

The order matters in one place: the record happens after the finalize, not
before. A flush that pads a chunk it then fails to deliver has not finalized it,
and subtracting padding from chunks that are full would hide data — the very
thing being fixed. Recording nothing on that path leaves the previous finalize's
record standing, which is still the last finalized chunk.

## Adjacent problem, not fixed here

The tests pin down two things worth knowing, both older than this change:

- A padded flush does not advance the append cursor, so the next append resumes
  inside the same chunk. The plane appended after the flush in the padded test
  case lands at append position 7, not 1.
- Nothing clears the chunk pool behind a flush, so the next flush re-sends the
  chunk it padded. The plane from before the flush therefore appears twice, at
  positions 0 and 6.

Neither loses data and both are inside the extent this PR publishes, so a reader
sees every frame. But a frame at a shifted position, and a duplicate of it, are
not what a caller appending after a flush would expect. Worth its own issue if
you agree.

## Verified

32/32 on the CPU-only build, including the new cases. The GPU config builds
clean (nvcc 13.1, sm_89) and registers `test_shape_after_flush_gpu`, but I did
not run it — no GPU in this session, so the GPU cases are on CI here.

nclack:
## gpu-tests coverage problem

Both `gpu-tests` legs finish in ~44 s and run **45** tests. This tree registers
54. Their list includes `test_aggregate` and `test_flush_drain_multiarray_gpu`,
two targets that left `tests/CMakeLists.txt` in #125 and #118 — and it does not
include either test added here.

The job runs `docker compose run --rm test`, which builds the image only when one
is missing. The self-hosted runner already has it, so `COPY . .` and the build
inside the Dockerfile have not re-run since the image was first built, around
2026-04-26. Both matrix legs share one image name, so the CUDA 12 / CUDA 13
split is not exercised either.

So the GPU coverage on this PR is: it compiles (checked locally, nvcc 13.1,
sm_89, and `test_shape_after_flush_gpu` links and registers), and the GPU cases
have not run anywhere. `cpu-tests`, `macos-tests`, `windows-tests` and `coverage`
build from source and do cover the change.

A `docker compose build` step, or `docker compose run --build`, would fix the
job. Left out of this PR — the first honest GPU run will take far longer than
44 s and may surface unrelated failures, which belong in their own change.

nclack:
Ran the suite on an L40 (sm_89, driver 580.126.20, nvcc 13.1) at 7c4887d, since CI could not: **52/52 passed**, `-E "(s3)"` for the two tests that need object storage. That includes `test_shape_after_flush_gpu`, so the GPU backend publishes the same extents as the CPU one — 8 for both cases in #193, where main publishes 6 and 2.

nclack:
## Review pass

A high-effort review of this branch confirmed three things worth recording here.

**The failure path is a real limit, not an oversight.** The extent is recorded
after the finalize that closed the padded chunk succeeds. If a later step of the
same flush fails, the record never happens and the published extent counts that
chunk's padding as data — up to `chunk_size - 1` planes. `main` avoided this by
bounding the extent to the appended element count, which is exactly what hides
frames after a successful flush. Both are wrong in one direction; this PR picks
the direction that never hides data.

**Moving the record inside the finalize does not work as a small change.** I
tried it: arm the padding before the flush delivers anything, and have whichever
finalize closes the chunk record it in the same step. That removes the failure
window, but `deliver_to_shards_batch` cannot tell which chunk is the padded one.
Gating on "last chunk of the batch" is right for a single-level stream and wrong
for an append-downsampled level whose padded chunk arrives in the later
append-drain batch: the padding lands on a chunk that is full, and the level's
extent then hides real planes on the success path. That is worse than the
failure-path over-report, so I reverted it. A correct version needs the flush to
tell each level which chunk carries its padding, which is more than this PR
should carry.

**A declared-size clamp is not the fix for bounded append dims.** After a
mid-stream flush a bounded dim can publish past its declared size. Clamping to
that size truncates frames the writer legitimately accepted, because capacity is
the chunk-rounded declared size: with `t(size=10, chunk_size=4)` the writer takes
12 frames, and clamping to 10 makes frame 11 unaddressable. Reverted as well.

All three trace to the same root cause, which this PR does not fix: a padded
flush leaves the append cursor inside the chunk it closed, so later frames land
at shifted grid positions and the padded chunk gets re-sent. Worth its own issue.

## Verification

CPU-only build 32/32, and 52/52 on an L40 (sm_89, driver 580.126.20) including
`test_shape_after_flush_gpu`. The `gpu-tests` checks here are not evidence — see
the note above about the stale runner image, fixed separately in #196.

PR #198, opened 2026-08-10
GPU path cleanup after the scatter work

### LINKED ISSUE #195: GPU path cleanup left over from the scatter kernel work
Four things the GPU path carries that #191 either touched the twin of, or
deliberately left alone.

## An unreleased pool slot when a multiscale dispatch fails

`ingest_dispatch_multiscale` acquires the staging pool's consume generation, then
runs four failure points before releasing it: the epoch-bounds check, the
device-to-device copy, and both timing event records. Each jumps to `Error`
without `gpu_pool_release_consume`, so that slot's consume generation never
advances and a later drain waits on an edge whose event was never recorded.
#191 fixed exactly this in `ingest_dispatch_scatter`, one function above.

## One of twenty kernel launches checks whether it was refused

`src/gpu` launches kernels at 20 sites across `transpose.cu`, `lod.cu`,
`compress.cu` and `aggregate.cu`. Only `transpose.cu` reads
`cudaGetLastError()`. Everywhere else, a launch the driver turns away — a wrong
arch for the binary, a bad configuration — is reported nowhere: the surrounding
code is driver-API, which never sees the runtime's per-thread error, so the
pipeline compresses and writes chunk memory the kernel never touched.

The one site that does check has to clear the error first, because the other 19
leave theirs unread and the same host thread drives several of them. So the
check is sound only as long as it is the only one.

## Device copies no kernel reads

`aggregate_layout_upload` allocates and fills `d_lifted_shape` and
`d_lifted_strides`, `aggregate_layout_destroy` frees them, and
`aggregate_layout_device_bytes` charges the memory estimate for them. No kernel
reads either pointer. This is the same thing #191 removed from
`tile_stream_layout_gpu`, one file over.

## A scatter layout is refused mid-append, not at open

The scatter derives its epoch split from `epoch_elements` and the shape on every
dispatch, so a layout it cannot place is refused after frames have been
accepted rather than when the stream opens. No configuration reaches this today:
`dims_n_append` only extends the append prefix across dimensions whose chunk
size is 1, which is what makes the skipped dimensions harmless. It is reachable
by a future change to that function.

## Test layouts do not have production geometry

`build_lifted_layout` in `tests/index.ops.util.c` duplicates the core of
`compute_level_layout` but sets the chunk stride to the chunk element count,
with no codec alignment; production pads to the sink's alignment, 4096 since
 #180. So no GPU kernel test scatters with a padded chunk stride, and the
duplicate can drift from the function it stands in for. It also skews which
paths the tests reach: padded strides inflate destination offsets, so a real
stream crosses from the scatter's 32-bit index to its 64-bit one at sizes no
test approaches.

nclack:
Five things the GPU path carried after the scatter kernel work in #191,
and a context bug the rebuilt GPU CI turned up.

ingest_dispatch_multiscale held the staging pool's consume generation
across four failure points, so a failure there left that slot stuck and
a later drain waiting on an edge whose event was never recorded. The
copy and its two event records move into a helper that owns the timing
slot. The epoch-bounds check runs before the H2D copy, which it does not
need.

Kernel launches were checked at 1 of 20 sites. CUDA_LAUNCH wraps each
one and reads the runtime's per-thread error that the surrounding
driver-API calls never see. It takes the pending error first, since
nvcomp runs kernels on the same thread and a launch that succeeds leaves
an error already stored there for the check to find. The prefix sums in
the aggregate path report their own status now too.

aggregate_layout uploaded a device copy of its lifted shape and strides
that no kernel read. The two allocations, the two frees, the term in the
memory estimate and the two fields come out together, and with them the
three functions that existed only to manage them.

The scatter refused a layout it could not place on the dispatch that
reached it, after frames had been accepted. Every level's layout is
checked where the array is built instead, so a bad one fails at open.
The per-dispatch check stays for the callers that pass layouts of their
own.

The tests duplicated compute_level_layout but set the chunk stride to
the chunk element count, with no alignment, so no GPU kernel test
scattered with a padded stride. compute_level_layout is exported from
src/stream and the tests call it with an alignment.

A stream used from a thread other than the one that created it refused
its kernel launch. The launch goes through the runtime API, which uses
the calling thread's context and rejects a stream owned by another,
while the driver-API calls around it resolve the context from the handle
and carry on. Both stream types capture their context and make it
current at every entry point, and put the caller's back on return. This
fails on main too, on CUDA 12, since #196 made the GPU tests build from
source.

Closes #195

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -184,8 +184,13 @@ jobs:
     env:
       CUDA_VERSION: ${{ matrix.cuda.version }}
       CUDA_MAJOR: ${{ matrix.cuda.major }}
+      # One image and one set of containers per CUDA major version.
+      COMPOSE_PROJECT_NAME: chucky-cuda${{ matrix.cuda.major }}
     steps:
       - uses: actions/checkout@v4
+
+      - name: Build image (CUDA ${{ matrix.cuda.version }})
+        run: docker compose build test
 
       - name: Build and test (GPU, CUDA ${{ matrix.cuda.version }})
         run: docker compose run --rm test
--- a/src/gpu/aggregate.cu
+++ b/src/gpu/aggregate.cu
@@ -209,13 +209,13 @@ aggregate_batch_slot_init(struct aggregate_slot* slot,
      cuMemHostAlloc((void**)&slot->h_permuted_sizes, C * sizeof(size_t), 0));
 
   slot->temp_bytes = 0;
-  CHECK_SILENT(Error,
-               CUDA_CALL(cub::DeviceScan::ExclusiveSum(nullptr,
-                                                       slot->temp_bytes,
-                                                       slot->d_permuted_sizes,
-                                                       slot->d_offsets,
-                                                       (int)C,
-                                                       (cudaStream_t)0)) == 0);
+  CUDA_CALL_OR(Error,
+               cub::DeviceScan::ExclusiveSum(nullptr,
+                                             slot->temp_bytes,
+                                             slot->d_permuted_sizes,
+                                             slot->d_offsets,
+                                             (int)C,
+                                             (cudaStream_t)0));
 
   if (slot->temp_bytes > 0)
     CU(Error, cuMemAlloc((CUdeviceptr*)&slot->d_temp, slot->temp_bytes));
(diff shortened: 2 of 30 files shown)


nclack:
Verified on an L40 (sm_89) at 155ac24: 51/51 ctest with `-E "(s3)"`, and 31/31 in
the CPU-only build. Both builds are warning-free.

Two notes on choices the issue left open.

Exporting `compute_level_layout` cost gcc the proof that `rank > 0`, which it had
while the function was static, so it warned that `ts` and `chunk_count` may be
used uninitialized. The two rank checks at the top restore it and are worth having
on a function the tests now call directly.

The test alignment is 4096. The codec's real input alignment is 4 or 8 bytes,
which for these chunk sizes pads nothing, so the tests would still not see a
padded stride. A value wider than the test chunks is what makes the padding
visible.

nclack:
The rebuilt GPU image turned up a failure. Fixed here.

`test_flush_drain_gpu` failed on cuda12 with `invalid resource handle` at the
scatter launch. Main fails the same test the same way at `transpose.cu:184`
(run 31439828721, the merge of #196), using the launch check that was already
there — so this predates the PR, and the PR only changes the wording of the
message.

The test calls `writer_flush` from a worker thread, while the context was made
on the main thread with `cuCtxCreate`. Driver-API calls on that thread still
work, since they take the context from the handle they are given, but a kernel
launch goes through the runtime API, which binds the calling thread's own
context and refuses a stream that belongs to another one. `delivery_main`
already makes the context current on its worker; the entry points callers reach
did not.

So the stream captures its context at create and makes it current in append,
flush and destroy. The multiarray path gets the same treatment in `update_impl`,
`flush_impl` and its destroy — same hazard, and fixing only one of them would
leave the other waiting for a caller to flush off-thread.

cuda12 now passes 53/53 with `test_flush_drain_gpu` green; all six checks pass.
Local L40 runs (CUDA 13) do not exercise this — CUDA 13 tolerated the mismatch
before the fix, and I have not established why.

PR #199, opened 2026-08-10
Test timing metrics

### LINKED ISSUE #197: Timing metrics from CUDA events have no test coverage
No test asserts that a timing metric derived from CUDA events ever arrives.

Ten metrics are accumulated through `accumulate_metric_cu_if_ready`: the H2D
copy and the scatter (reported as `Copy` when multiscale is on), the four LOD
stages, compress, the tail gate, aggregate, and D2H. Each depends on a caller
recording two events and then marking a ring slot, and nothing checks that the
marking happens. The only metric any test looks at is `sink`
(`tests/test_flush_orchestration.c`), which is timed on the host through
`accumulate_metric_ms` and does not use a ring at all.

A dropped metric is close to invisible. `print_metric_row` in
`bench/bench_report.c` skips a metric whose count is zero, so the stage's row
disappears from the report rather than reading as a zero, and the JSON carries
the same empty numbers. Nothing fails, and no test covers the path.

#191 hit this. It moved `st->pending = 1` out of `scatter_timing_begin` and into
one of that function's two callers, so every measurement from
`ingest_dispatch_multiscale` was discarded. The build was warning-free and the
suite was 50/50 green across six runs on an L40, before and after. What found it
was reading the second caller by hand; what confirmed it was the missing `Copy`
row in a multiscale bench.

The two lost-sample counters, `scatter_samples_lost` and `lod_samples_lost`, are
also never asserted, so a ring that silently overwrites unread measurements
looks the same as one that does not.

nclack:
No test asserted that a metric derived from CUDA events ever arrives. A
dropped measurement leaves the count at zero, and the report drops a
zero-count row, so the stage vanishes instead of reading as zero. #191
lost every measurement from the multiscale ingest path that way and the
suite stayed green.

Each metric is now checked where it is produced. test_ingest covers the
H2D copy and the scatter on both dispatch paths. It orders their timing
events against a marker, so a stale event cannot pass as fresh, and
overruns the ring on purpose so the lost-sample counter reports a loss.
test_d2h_deliver covers compress, the tail gate, aggregate and D2H.
test_metrics is new and runs a whole stream in three geometries, the only
way to reach the four pyramid stages.

test_metric_check.h holds the shared checks: the count is the contract,
and an interval that brackets work must also show a positive duration.

ingest_destroy now clears the state it frees, so a failed setup followed
by the caller's teardown no longer frees the same handles twice.

Closes #197

### DIFF
--- a/src/gpu/stream.ingest.c
+++ b/src/gpu/stream.ingest.c
@@ -83,6 +83,9 @@ ingest_destroy(struct staging_state* stage)
     cu_event_destroy(stage->timing[i].t_start);
     cu_event_destroy(stage->timing[i].t_end);
   }
+  // Setup tears itself down before returning an error and the caller tears
+  // down again, so this has to be safe to run twice.
+  memset(stage, 0, sizeof(*stage));
 }
 
 void
--- a/src/gpu/stream.ingest.h
+++ b/src/gpu/stream.ingest.h
@@ -19,7 +19,9 @@ ingest_init(struct staging_state* stage,
             struct gpu_ordering* ord,
             CUstream compute);
 
-// Free staging buffers and events.
+// Free staging buffers and events. The caller synchronizes the streams first:
+// a copy still reading a staging buffer faults once it is freed. Safe to call
+// twice, and on a state whose setup failed.
 void
 ingest_destroy(struct staging_state* stage);
 
--- a/tests/CMakeLists.txt
+++ b/tests/CMakeLists.txt
@@ -200,5 +200,6 @@ add_test_exe(test_append_latency test_append_latency.c stream_config chucky_log)
 add_gpu_test_exe(test_lod         test_lod.c         LINKS CUDA::cuda_driver CUDA::cudart_static lod lod_plan index_ops morton_util)
 add_gpu_test_exe(test_lod_dtypes  test_lod_dtypes.cu LINKS CUDA::cuda_driver CUDA::cudart_static lod lod_plan)
 add_gpu_test_exe(test_multiscale  test_multiscale.c  LINKS CUDA::cuda_driver CUDA::cudart_static stream Zstd::Zstd test_data test_shard_sink)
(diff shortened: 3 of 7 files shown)


nclack:
Superseded by the verification below. This described the first round of tests,
before the review passes and before #198 merged.

nclack:
Superseded. This described the branch when it sat on `90b2f6a`. It has since
been merged with main for #198, which changed the ring-mark shape described
here and fixed the cuda12 failure. The `chunk_count` warning is gone too.

nclack:
#198 has since merged, so the cuda12 failure described above is fixed on main.

nclack:
Checked that these tests fail when the behaviour they name is broken, rather
than passing because nothing is wrong. Four faults reintroduced one at a time
on an L40, re-run after #198 merged:

| Fault | Caught by |
| --- | --- |
| The scatter ring mark never set | `test_ingest`, `test_metrics` |
| The scatter interval's end never recorded | `test_ingest`, `test_metrics` |
| A reused staging slot keeps a stale H2D start | `test_ingest` |
| The H2D start never recorded at all | `test_ingest` |

`test_d2h_deliver` and `test_flush_orch` stay green through all four. Before
the review the third fault passed everywhere; ordering the start events against
a marker is what closed it.

The `ms > 0` question was settled by measuring. The smallest single ingest
interval over 25 runs was 6.3 us for the copy and 7.0 us for the scatter,
against a timer resolving to about 0.5 us, so a zero there is a defect and not
a fast machine. Where the schedule decides how many measurements arrive, only
the total is checked.

52/52 pass and the build is warning-free, with the three metric tests repeated
ten times per run.

#198 paired the scatter record with its ring mark in one helper both dispatch
paths call, so the #191 shape — one caller forgetting the mark — can no longer
be written. These tests still cover stale and reused events, which the pairing
does not address.

Two gaps left deliberately. The flush stages seed their timing events the way
ingest does, so a cycle that skipped a record still reports a plausible
interval; closing that means extending the marker ordering to three more events
across two streams. And `d2h_deliver_destroy` has the same double-teardown
shape `ingest_destroy` just had, which this branch does not touch.

PR #200, opened 2026-08-11
Add an L40 sweep at 7af733d

nclack:
Add a compress and lod sweep on an L40 at 7af733d. All 204 runs passed.

Throughput is unchanged against the full sweep at 170bfcf. The median run
moves +0.3%, and every scenario, codec and backend group stays within 2%.
Only smallepoch swings more, +26% on lz4 at 64K, and it measures a single
append, so one sample decides the number.

The scatter kernel is 22-40% faster on the single-scenario runs, the win
from #191. It falls from 8-11% of wall time to 5-7%, too small a share to
show up end to end. On the multiscale runs scatter is already three times
cheaper per call and is unchanged.

Both tiers use fill=xor and sink=discard, so the runs line up with the
170bfcf file. The CPU runs used 16 cores, as the 170bfcf sweep did. No
results file records the core count, and CPU throughput scales with it.


nclack:
Tier choice: #199 is a test PR whose only source change is a `memset` in
`ingest_destroy`, so it cannot move throughput. What this sweep measures is
everything since the 170bfcf baseline: #188, #191, #194, #198, #199. compress
and lod cover the GPU single-stream grid and the multiscale grid, which is
where the scatter kernel #191 and #198 touched does its work. The io, fill and
s3 tiers use a different sink or fill, so they would not compare against the
baseline run-for-run.

Verification: 53/53 ctests pass at 7af733d on the L40. `compression_fold`
matches the baseline on all 204 runs, so both sweeps did identical work.
Per-stage medians are flat except scatter: memcpy -5%, h2d -1 to -3%, d2h +2
to +4%, compress -0.5%, and the four pyramid stages within 1%.

The core count matters more than expected. The first run of this sweep used
an 8-core allocation and read as a uniform 44% regression across all 72 CPU
multiscale runs. That was the allocation, not the code: `init_s`, the bench
data generation that no commit since the baseline touches, moved by the same
1.97x factor, and only on CPU runs (GPU runs were 1.01x). An L40 node has 128
logical CPUs, and `--max-threads` defaults to the OpenMP default, so the CPU
numbers follow whatever Slurm handed out. A 16-core allocation reproduces the
baseline within 1% on two spot-checked runs, 0.773 vs 0.766 GiB/s and 0.946
vs 0.950, which is why this file ran on 16.

Worth a follow-up issue: record the core count in the results file, or have
the sweep pass an explicit `--max-threads`. Right now two sweeps' CPU numbers
are only comparable if the two allocations happened to match, and nothing in
the file says whether they did.

nclack:
Correction to the comment above, since it changes what the fix should be. I
wrote that `--max-threads` defaults to the OpenMP default. It does not — that
is what the comment in `bench_util.h` says, but the code path calls
`platform_default_thread_count()`, which returns
`sysconf(_SC_NPROCESSORS_ONLN)`: the CPUs online on the whole node, ignoring
the cgroup the job runs in. On a `-c 16` job here sysconf reports 192 while
`sched_getaffinity` reports 16.

So the CPU numbers do not vary because the thread count follows the
allocation. They vary because it does not. `src/cpu/stream.c` and
`src/multiarray/stream.c` size their pool at `nthreads - 1`, so a 16-CPU
allocation runs 191 workers on 16 CPUs. `src/gpu/stream.init.c` caps its copy
helpers at 3, which is why the GPU runs in this sweep were insensitive to the
allocation and the CPU runs were not.

The default is slower than any explicit setting at or below the core count. On
a `-c 64` job, orca2_multiscale zstd 256K: default 1.65 GiB/s,
`--max-threads 32` 1.84, `--max-threads 64` 2.06.

That makes it a defect rather than a benchmark-metadata problem, and it
applies anywhere the process is CPU-limited, including a container. Using
`sched_getaffinity` on Linux (and the cgroup quota when one is set) would fix
both the sweep comparability and the deployment case.

nclack:
outdated

PR #203, opened 2026-08-12
Count pending bytes inside the io queue

### LINKED ISSUE #201: Filesystem sink can report a huge pending-byte count
`shard_pool_fs` tracks bytes waiting to be written with two counters: `queued_bytes`, added to by the thread that hands a write to the queue, and `retired_bytes`, added to by the worker once the write finishes. Both write paths post the job before adding to `queued_bytes` — `fs_slot_write` at src/zarr/shard_pool_fs.c:106-110 and `fs_slot_write_direct` at src/zarr/shard_pool_fs.c:163-167.

So the worker can pop the job, finish the write, and add to `retired_bytes` (src/zarr/shard_pool_fs.c:67 and :139) before the posting thread reaches its `atomic_fetch_add`. In that window `pool_fs_pending_bytes` computes `queued_bytes - retired_bytes` and returns it as `size_t` (src/zarr/shard_pool_fs.c:338-343), so the result wraps to a number near 2^64. Small writes on a fast drive make the window easy to hit.

Cost today:

- Reporting. The bench reads pending bytes just before flush and divides by flush time to print a flush rate (bench/bench_util.c:513). The GPU writer keeps a high-water mark (src/gpu/stream.c:172-174) that ships as `peak_pending_mib` in the results JSON and shows on the sweep pages (scripts/sweep/overview.html:203). One wrapped sample ruins the mark for the whole run.
- Stalls, if anyone enables them. `apply_backpressure` (src/gpu/stream.c:170-201, called from :280) is the only control that acts on the number, and a wrapped sample drops it into the drain loop until the 30 s timeout. It costs nothing today: `backpressure_bytes` defaults to 0, meaning off (bench/bench_util.c:657, src/types.stream.h:109), no recorded sweep passes `--backpressure`, and the CPU writer never reads pending bytes. Turning it on turns this into a 30-second stall.

No stored sweep shows a wrapped value — the largest recorded `peak_pending_mib` is about 1.6 GiB — but the mark is sampled only once per staging buffer handed to the device, so samples are sparse.

This blocks trusting pending-byte measurements for #178, which cites peak pending (0.8-1.25 GiB) as its evidence.

nclack:
shard_pool_fs reports the bytes its queued writes are still carrying: an
upper bound, since a write counts from the moment it is accepted, so the
figure can read high but never low. It used to be a subtraction of two
counters, one raised by the thread queueing a write and one by the worker
that finished it. Two separate reads are not a consistent snapshot, and
the result was unsigned, so it could wrap to a number near 2^64. The pool
also counted a write only after handing it to the worker, so the worker
could finish and lower the total before the raise landed.

The io queue keeps the count now. io_queue_post_bytes takes a byte count
and raises the total inside the same lock that puts the job in the queue;
the worker lowers it inside the same lock that marks the job finished. A
reader can never see less work than the queue holds, so no ordering rule
is left to break. io_queue_post is that call with a count of zero, leaving
jobs that carry no data unchanged.

shard_pool_fs and the throttled bench sink drop their own counters and read
the queue's instead. The shard_pool and shard_sink headers state the bound,
and pending_bytes is uint64_t rather than size_t throughout: both
interfaces, every implementation behind them, peak_pending_bytes, and
backpressure_bytes.

tests/test_io_queue.c covers the count directly. It blocks the worker,
checks the total after each post, then checks that it returns to zero once
the queue drains and that a job posted without a byte count never changes
it. The pool test checks that both write paths, copying and zero-copy,
report through. Tests now share one wait-for-flag helper instead of four
copies.

Closes #201.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,6 +1,6 @@
 target
 build
-build-cpu
+build-*
 result
 .ninja_log
 .clangd
--- a/bench/bench_report.c
+++ b/bench/bench_report.c
@@ -105,7 +105,7 @@ print_bench_report(const struct stream_metrics* metrics,
                    float wall_s,
                    float init_s,
                    float flush_s,
-                   size_t flush_pending_bytes)
+                   uint64_t flush_pending_bytes)
 {
   const size_t chunk_bytes = layout->chunk_stride * dtype_bpe(dtype);
   const size_t num_epochs =
@@ -182,7 +182,7 @@ print_bench_report(const struct stream_metrics* metrics,
                    (unsigned long long)metrics->lod_samples_lost);
     print_append_latency(metrics);
     char pbuf[32];
-    format_bytes(pbuf, sizeof(pbuf), (uint64_t)metrics->peak_pending_bytes);
+    format_bytes(pbuf, sizeof(pbuf), metrics->peak_pending_bytes);
     print_report("  peak pending:    %s", pbuf);
   }
 
--- a/bench/bench_report.h
+++ b/bench/bench_report.h
@@ -44,7 +44,7 @@ print_bench_report(const struct stream_metrics* metrics,
                    float wall_s,
                    float init_s,
                    float flush_s,
-                   size_t flush_pending_bytes);
+                   uint64_t flush_pending_bytes);
 
 // Emit the pass-case JSON report to stdout. sink_metric may be NULL (no sink
 // block is written in that case).
--- a/bench/bench_util.c
+++ b/bench/bench_util.c
@@ -510,7 +510,7 @@ run_bench(const struct bench_config* cfg)
                                   dtype_bpe(dtype),
                                   cfg->append_elements) == 0);
 
-  size_t pending_bytes = bench_zarr_pending_bytes(&zarr);
+  uint64_t pending_bytes = bench_zarr_pending_bytes(&zarr);
 
   struct platform_clock flush_clock = { 0 };
   platform_toc(&flush_clock);
@@ -1066,7 +1066,7 @@ run_bench_two_streams(const struct bench_config* cfg)
       print_metric_row(&m[k].backpressure);
       print_append_latency(&m[k]);
       char pbuf[32];
-      format_bytes(pbuf, sizeof(pbuf), (uint64_t)m[k].peak_pending_bytes);
+      format_bytes(pbuf, sizeof(pbuf), m[k].peak_pending_bytes);
       print_report("  peak pending:    %s", pbuf);
     }
   }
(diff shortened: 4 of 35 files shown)


nclack:
Verification, CPU-only build on a `cpu` node, RelWithDebInfo, no warnings:
33/33 ctest, and the new test 10/10 over repeats.

The test earns its keep by mutation. Reverting to the old order — post the
job, then count it — fails it 5/5 runs, reporting `pending while parked:
18446744073709547520 bytes (queueing 4096)`, which is 2^64 - 4096.

Two things worth a second opinion.

The first test I wrote sampled `pending_bytes` from one thread while
another queued small writes, first with one writer and then with 24 on 16
cores. It passed against both broken builds, so I threw it away. The wrap
needs the difference to be near zero at the moment of an uncounted post,
and load keeps the queue deep, so hammering the pool makes the bug less
reachable, not more. The window is a few nanoseconds wide and needs a
thread descheduled inside it, which is why the hook exists. That also
softens the issue's claim that small writes on a fast drive make the
window easy to hit — it needs bad luck at one instruction, which sparse
sampling over hours of sweeps can still find.

The load-order half is reasoned, not tested. Covering it would need a
second hook inside `pool_fs_pending_bytes`, which the GPU producer calls
per staging buffer, and that seemed steep for a reporting counter. Happy
to add it if you disagree.

Nothing outside `pool_fs_pending_bytes` reads `queued_bytes` or
`retired_bytes`, so collapsing them into one signed counter would delete
this bug class outright, with no two-load question and one atomic read.
I kept the two counters because the issue named that fix; say the word and
I will swap it.

nclack:
Switched to the single counter, so the two ordering rules the earlier
version depended on are gone. Verification on a `cpu` node, CPU-only
RelWithDebInfo, no warnings: 33/33 ctest, new test 10/10 over repeats.

Mutation results, all rebuilt and rerun five times each:

| mutation | outcome | pending while parked |
| --- | --- | --- |
| count after the handoff | fails 5/5 | 0 |
| drop the negative clamp | passes 5/5 | 4096 |
| both | fails 5/5 | 18446744073709547520 |

Worth knowing why the assertion is equality rather than an upper bound.
With one signed counter, counting after the handoff drives it to -4096 and
the clamp reports that as 0, which an upper-bound check accepts — the first
version of this test passed against that build. The middle row is there to
show the clamp is belt-and-braces: on a correct tree the counter never goes
negative, so removing it changes nothing.

The bench sink keeps its two counters. Its read already loads retired first
and clamps, so it under-reported as 0 instead of wrapping, and converting
it is a bigger change than the ordering fix it needed.

nclack:
Verified on a `cpu` node and an L40: 33/33 CPU ctest, 54/54 GPU, both
warning-free, and 8/8 CI.

Mutation check on the queue's accounting, each rebuilt and rerun five times.
Dropping the raise fails both the queue test and the pool test; so does
dropping the lower.

The two comments above describe earlier versions of this branch — first a
two-counter reorder, then a signed counter with a clamp. Neither survives.

PR #204, opened 2026-08-13
Record memory and run details in sweeps

### LINKED ISSUE #64: benchmark sweep: record memory usage
actual relative to a baseline, and estimated

### LINKED ISSUE #202: add details to sweep results
A results file records five things about the run that produced it: name, hostname, gpu, commit and date (the `machine` block in `scripts/sweep/sweep.py`). Should add:

* thread count
* build configuration
* driver and CUDA version
* frame count per scenario

nclack:
Closes #64. Closes #202.

Each run in a sweep now records its frame count, the size of the thread pool
the stream built, and the memory it took: resident memory before the stream
existed, peak resident memory over the run, device memory, and which of those
the estimate can be held against. The engine's own estimate was already
recorded, so an estimate can now be checked against a measurement.

Each sweep now records the driver version, the number of cores the sweep was
given, and a build block read from the build directory: build type, CUDA
architectures, C++ compiler, nvcomp path, and CUDA compiler version.

The explorer shows the sweep detail under the machine line and the run detail
in the tooltip. A sweep written before a field existed says unknown.

platform_resident_memory and platform_peak_resident_memory are new, on all
three platforms.

Writing those details down exposed two bugs, fixed here. The CUDA architecture
default was set after enable_language(CUDA) had already cached the compiler's
own, so it never applied and a plain configure built for sm_75. The CPU
pipeline pool counted the machine's cores rather than the ones the process was
allowed, so a sweep under a batch scheduler started a thread per core and then
shared a fraction of them. Results files move to version 6, because no
CPU-backend timing compares across that change.


nclack:
No schema version bump. Every field is additive, and the README rule is that
adding one does not need a bump; a bump would say a metric changed meaning,
which is how `retired_metrics` reads it. #202 asked for one — say the word and
I will add version 6 with a no-op migration.

Verification so far: the new platform readings pass a test that fails when the
kilobyte-to-byte conversion is dropped, checked by mutation. `report.py`
regenerates the site from all 25 stored sweeps, which lack the new fields, and
from a synthetic file that has them. The benchmark itself has not been built or
run on hardware yet — CI covers the build, and a sweep would confirm the
numbers are sane.

nclack:
All six checks pass. Interesting notes:

- clang at -O2 deleted the 64 MiB allocation the test measures, so macOS saw no
  growth. The test now writes one byte per page through a volatile pointer.
- macOS keeps the pages after `free`, so the next test started from an already
  inflated baseline. Linux hands them back and hid this.
- Linux `ru_maxrss` trailed `/proc/self/statm` by 90 KB in the container. The
  two readings come from different accounting in the kernel, so the test checks
  each against the block instead of against the other.

The peak on macOS now comes from `task_info` rather than `getrusage`, so both
readings there are in bytes from one source.

nclack:
On the two bug fixes, and a correction to my earlier comment: there **is** a
schema bump now, to 6. Not for the added fields, which are still additive, but
because the CPU thread fix moves every CPU-backend timing.

**CUDA architectures.** `enable_language(CUDA)` caches the compiler's own
default, so the `if(NOT DEFINED ...)` guard below it never fired and `89 100`
was dead. Verified by configuring: before the fix the cache held
`CMAKE_CUDA_ARCHITECTURES:STRING=75`, after it holds `89;100`, and
`-DCMAKE_CUDA_ARCHITECTURES=89` still wins. It is a cache entry now, so the
sweep can read the value back — a plain `set()` before `enable_language` works
but leaves nothing in `CMakeCache.txt`.

Worth knowing: the GPU runner is an RTX 5070, which is sm_120, so it matched
neither the old default nor the new one. A bare architecture number also emits
PTX, which is why both work — the runner JITs at load. Both GPU jobs pass, and
the build costs about ten seconds more for the second architecture. If you want
the runner covered natively, add 120 to the list.

**CPU pool.** `platform_default_thread_count` used `_SC_NPROCESSORS_ONLN`,
which ignores the affinity mask. Under `srun -c 16` on a 96-core node the
pipeline pool started 96 threads. `tests/test_platform_threads.c` narrows its
own mask to one core and asserts the count follows; reverting the fix makes it
report 96 against 1 allowed core, so the test does hold the behavior down.

Every stored sweep still migrates and validates, and the site regenerates from
all 25 of them.

PR #206, opened 2026-08-14
Use s3-blackhole for S3 benchmarks

### LINKED ISSUE #205: use s3-blackhole for s3 benchmarks instead of minio
Currently, when benchmarking s3 write speeds, I spin up minio in a docker container.

I'd like to replace minio with s3-blackhole for this. `minio` is still useful for testing/validation. This is purely about benchmarking.

[s3-blackhole](https://github.com/nclack/s3-blackhole) is a simple s3 server I wrote that exposes some minimal set of endpoints to test s3 write speeds. It does not validate or store any data, and instead simply discards anything it receives.

As a result, any benchmark measurements I make using this server should be producer-limited.

nclack:
Closes #205.

## What changed

- Add a small multi-stage image that builds a pinned `s3-blackhole` revision.
- Expose it on localhost through an opt-in `benchmark` Compose profile with a health check.
- Add `--backend cpu|gpu` so a sweep can target one backend while preserving the existing mixed-backend default.
- Document the CPU-only S3 sweep start, run, inspect, and stop workflow.

MinIO remains unchanged as the backend for integration tests that read data back and validate it.

## Why

MinIO persists uploads, so its storage path can limit S3 write measurements. `s3-blackhole` drains complete S3 uploads and discards them, keeping these benchmarks focused on producer throughput.

The S3 matrix previously hard-coded both CPU and GPU runs with no CLI filter. The selector lets a CPU-only machine run the 96 applicable configurations without also recording 48 expected GPU failures.

## Validation

- `docker compose config --quiet`
- `docker compose --profile benchmark config --quiet`
- `docker compose up --build --wait s3-blackhole`
- Ran `bench_stream_256cube_single` with the CPU backend against the service: 96 MiB uploaded successfully through 16 multipart uploads, with no aborts, unsupported requests, or body-read errors reported by `s3-blackhole`.
- S3 dry-run matrices: `--backend cpu` selects 96 runs, `--backend gpu` selects 48, and omitting `--backend` preserves all 144.
- `--backend tpu` exits with Click's expected invalid-choice error.


nclack:
Added `--shards 4` to the server's command, and dropped the commit pin so the image builds the default branch.

s3-blackhole now defaults to one accept thread per core. That is wrong here, because the server shares a machine with the sweep and every accept thread holds a core the sweep could be sending from. Measured on a 64-core node with `256cube_single`, no compression, 37.5 GiB, five interleaved repeats, medians of client GiB/s:

| Server | GiB/s |
| --- | ---: |
| `3b736b6`, the old pin (single-threaded) | 6.57 |
| default branch, one thread per core (64) | 5.42 |
| default branch, `--shards 4` | 6.91 |

At its own default the newer server is slower than the commit this branch used to pin. With four threads it is faster than either.

A shard sweep on that node peaks at two threads (7.01) and falls to a floor from sixteen on (5.4). The same client across two nodes shows no falloff — 7.8 to 8.1 GiB/s everywhere from two threads up — so the cost is the accept threads taking cores from the client, not the receive path. Nothing is given up by fixing it at four.

Verified the unpinned `cargo install` line resolves and builds (currently `793ab5b5`), the server starts with the new command and reports 4 accept threads, and `/healthz` returns 200. Not built as a Docker image — no Docker on the cluster node I measured from.

The builder image moved from `rust:1.85-bookworm` to `rust:1-bookworm`, since an unpinned source with a toolchain pinned at the declared minimum will break the first time upstream raises it.

nclack:
Added an S3 baseline measured against the server this branch now configures: `bench/results/reef-l40-e365889-20260819.json`.

- One CoreWeave L40, `--qos dev -c 16 --gres gpu:l40:1`, GPU build, `-DCMAKE_CUDA_ARCHITECTURES=89`, warning-free
- Full `s3` tier, both backends, 144 runs, all pass (the laptop file is 96, CPU-only)
- s3-blackhole `793ab5b` run natively with `--shards 4`, matching the image's command; no Docker on the cluster
- Server received 130.5 GB over 29,958 requests with 0 body read errors, and reported 4 accept threads
- fs-sink scratch on the node's local `/tmp`

Median GiB/s in, by sink and backend:

| Sink | CPU | GPU |
| --- | ---: | ---: |
| discard | 3.73 | 10.52 |
| fs | 2.21 | 2.46 |
| s3 | 2.39 | 4.38 |

The S3 sink now costs about what the local filesystem does, and is the limit for GPU runs rather than a wall. Its stage is 36% of wall on `256cube_single__none__xor__cpu__u16__2M__s3__10gbps`, against 77% in the laptop file.

**The two files are not comparable machine to machine.** On the 96 shared ids the s3 runs are 4.8x faster here, but `fs` is 0.56x and `discard` is 0.61x — those sinks have nothing to do with this change, so the machine alone moved things in both directions. Do not read the 4.8x as the effect of the server change; the honest comparison is s3 against fs and discard within one file. Note also that CPU-backend numbers track the core count of the allocation, which the schema does not record: this ran at `-c 16`, and the server's four accept threads shared those cores.

I left `laptop-s1qrti9n-16703f3-20260814.json` in place, since it is still a real measurement of a real machine and the report keeps old sweeps. Happy to drop it if you would rather the branch carry only a current baseline — it measured the pre-optimization single-threaded server behind a published Docker port, so nothing in it describes what this branch now sets up.

One caveat worth recording: this file is schema 5. `#204` on main moves sweeps to schema 6 with memory and run details, so re-running after this branch catches up to main would produce a more complete record.

PR #207, opened 2026-08-19
Break the trend chart out by input

nclack:
Input becomes a dimension of its own on the overview page. The Trend card
draws one panel per input over a shared date axis, with a Same scale
tickbox that puts every panel on one value axis. An Input filter picks the
one the machine cards and the latest sweep use, and clicking a panel title
sets it. The trend table and the biggest changes cover every input and
name it in a column.

The biggest changes now pairs runs by id against the last sweep that ran
the same configuration, so a configuration the sweep before skipped no
longer drops out.

### DIFF
--- a/scripts/sweep/README.md
+++ b/scripts/sweep/README.md
@@ -84,7 +84,9 @@ ten, `rand` barely at all. One line over all three jumps whenever a sweep adds o
 drops an easier input, so the trend chart draws a panel per input. **Same scale**
 puts every panel on one value axis, so their heights compare directly; untick it
 to read a panel whose numbers are much smaller. The **Input** filter picks the one
-the machine cards and tables use, and clicking a panel title sets it.
+the machine cards and the latest sweep use, and clicking a panel title sets it.
+The trend table and the biggest changes cover every input, with a column saying
+which one a row came from.
 
 ## What the numbers mean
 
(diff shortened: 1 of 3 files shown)


nclack:
Why: the trend line took the best run per sweep across every input, so the
zeros runs won whenever a sweep included the fill tier and the line dipped
when it did not. With `256cube_single` / zstd / gpu / discard on livescreen,
the old chart and machine card read 13.0 GiB/s from a zeros run; the xor
panel reads 7.96. The fill tier only ran on `orca2_single` and
`256cube_single`, which is why the jumps looked like regressions.

Verification: rendered both the pre- and post-change site from
`bench/results/` in headless chromium and drove the page. Checked three
panels with the expected point counts and legends, no console errors,
the crosshair tooltip naming its input, panel titles setting the filter,
the empty state still offering `... · xor` as a way out, a scenario the
fill tier never ran collapsing to one full-width panel, the trend table
carrying its new Input column across 36 rows, and light, dark, 620px and
900px layouts.

Two things worth knowing for review. Panels are drawn in a second pass
because the first panel measures the whole grid row until its siblings
exist, which sized its chart three times too wide. Same scale defaults to
on, which is right for throughput but useless for compression ratio, where
zeros reaches 5047x and rand 2.33x — hence the toggle.

Issue #178, opened 2026-08-08
Filesystem sink serializes writes through one worker


nclack:
The filesystem shard pool shares one `io_queue` and one worker across every
shard and level. Each job does a blocking `pwrite` (Linux, macOS) or a
synchronous `WriteFile` (Windows), so however deep the backlog gets, only one
write is ever outstanding. Filesystem bandwidth goes unused on every platform,
and splitting work across shards and levels buys nothing.

## Steps

- [x] Measure the write path — #208, merged as `fee70ed`
- [x] #212 — correctness at queue depth one, merged as `54b03fa`
- [x] Step 3, more queue depth — PR #224, which carries #213, #225 and #227
  - [x] #226 — the sweep records a depth in #224; the report axis is left
  - [x] #228 — publish a queue-depth sweep
- [x] #229 — close the gaps in the backend interface
- [ ] #214 — io_uring backend on Linux
- [ ] #215 — overlapped writes on Windows

Each one is its own pull request off `main`. The plan is `dev/io-scheduler.md`
in the repo; this list only tracks state.

Spun out: #211, plate fields share pool slots. Unrelated to the io queue.

Spun out of step 2, in the order they should be taken:

- #217, a fence waiter can be left holding a freed lock. Fixed inside #216 and
  closed with it.
- #219, the filesystem fault hooks are compiled into the shipping backend.
  PR #223, and the first of the three open pull requests to merge: it takes the
  noop hooks out of the shipping backend, while #222 moves a truncate hook the
  other way through the same file. See #231.
- #218, a flush misses errors from the truncates and closes it queued itself.
  PR #222, second to merge, so its truncate hook has the test-side wrapper #223
  adds to go into.

Spun out of those three, and depending on both of them:

- #231, the one-shot fault flags collapse into one pair of fields.
- #230, nine tests hand-build the same store, pool, array and stream.

## Evidence

In the `oreb` sweep, GPU `CODEC_NONE` filesystem output reached 1.72-2.18 GiB/s
(1.84-2.34 GB/s) while pending output peaked at 0.8-1.25 GiB. The target drive
sustains 5.94 GB/s at Seq1M Q8T1. Multi-shard writes are interleaved in FIFO
submission order, then run one after another.

Queue depth is not the only cause. On xfs a write that extends a file takes the
inode lock for itself, so writes to one shard serialize inside the filesystem
however many the application keeps outstanding. And `smallepoch_single` holds
[...]

PR #208, opened 2026-08-20
Measure the filesystem write path

nclack:
Only one write at a time is run by the filesystem sink. Measured first, before
any change to that.

Recorded now: how many distinct shard files have a write waiting, as a peak and
as a time-weighted average. That is the queue depth available; the depth
achieved is one by construction and is not recorded. Also recorded are shard
files opened and open at once, wait and run time per write, copied against
borrowed bytes, backlog peaks, and a request-size histogram. All of it is
written to the results file, the overview page and the explorer.

Counting per file requires knowing which file each write was issued for, so
each open of a shard file is now numbered rather than identified by a
descriptor that may since have been closed and reissued. The two post
functions are collapsed into one `io_queue_post(q, struct io_work)`. The
counters, being platform-independent, are shared from one file by both queue
implementations, and the measurement types are kept out of the installed
headers.

Added `smallepoch_4shards`, the same chunk over four shard files where its twin
holds one open. Both smallepoch scenarios are now in the sweep's I/O tier.

No behavior changes.

See #178


nclack:
## Adding smallepch_4shards scenario

In `smallepoch_single`, only has one shard file open at a time. That's useful
for measuring effects that manifest when we don't have any concurrency across
shards. Importantly, the one-file case is the only place pre-sizing can be 
measured, and pre-sizing is the only lever that shape has. 

But it's not the fastest for these kinds of high-framerate small-image
writes. 

In `smallepoch_4shards`, y and x are split in two, giving four. Measured on a
CPU node writing to local `/tmp`:

| scenario | files with a write waiting | files opened | open at once |
|---|---|---|---|
| `smallepoch_single` | 0.78 avg, 2 peak | 4 | 3 |
| `smallepoch_4shards` | 3.36 avg, 5 peak | 16 | 8 |

## Results

Open handles outnumber the shards. The single-shard scenario should hold one
shard file at a time and shows three open at once; the four-shard one shows
eight. A close is only queued, never waited on, so descriptors outlive the slot
that owned them. Worth knowing before writes start retiring out of order.

## Reading the numbers

Measurement runs from the first queued job to the end of the flush, so every
shard's closing footer falls inside it. Truncate and close carry no payload and
are not counted toward the files-waiting figures; the footer write is. A peak
can therefore come from the flush rather than from streaming. Compare runs on
the time-weighted average and treat the peak as an upper bound.

## Checks

`test_platform_memory` is flaky here and on `main` (1/20 and 4/20 over 20
runs). Filed as #209; unrelated to this branch.

README.md
```diff
@@ -165,6 +165,12 @@ multiscale, and multiscale-with-dim0-downsampling modes.
 
 - `bench_stream_256cube_single` / `_multiscale` / `_multiscale_dim0`
 - `bench_stream_orca2_single` / `_multiscale` / `_multiscale_dim0`
+- `bench_stream_medfmt_single` / `_multiscale` / `_multiscale_dim0`
+- `bench_stream_smallepoch_single` — a tiny epoch, one shard file at a time
+- `bench_stream_smallepoch_4shards` — the same chunk over four shard files,
+  so writes to different files can be overlapped
```
nclack:
```suggestion
- `bench_stream_smallepoch_4shards` — a tiny epoch, 4 concurrent shards
```

PR #210, opened 2026-08-21
Fix platform_resident_memory failure handling

### LINKED ISSUE #209: test_platform_memory fails intermittently on Linux 6.8
`test-test_platform_memory` fails intermittently on a CoreWeave `cpu` node
(kernel 6.8.12, Ubuntu 22.04). Over 20 runs: 4 failures on `main`, 1 on a
feature branch. It passes in CI, so this is not a hard break.

The failing assertion is the first one in the binary:

```
tests/test_platform_memory.c:30: check failed: resident > 0
```

Later tests in the same run read resident memory fine — one reports 1822720
bytes a few milliseconds afterwards — so the process does have resident pages
by then.

The read is not failing. `/proc/self/statm` reports resident 0 for a young
process on this kernel and settles a few tens of milliseconds later:

```
immediately:        0
after 4 more reads: 0
after touching 1MB: 0
after 20ms sleep:   385
```

Across 500 fresh processes on that node the first read was 0 every time, and
the value it settles to is correct. Touching pages does not bring it forward;
only elapsed time does. This looks like the per-CPU batched RSS counters,
which a young process can observe before the first flush.

`platform_resident_memory` returns 0 both when the read fails and when the
kernel reports 0. `test_resident_memory_is_plausible` asserts `resident > 0`
as a proxy for the read having succeeded, and on a kernel that batches those
counters the proxy does not hold, so the test races process startup. Whether
it fails depends on how much time passes before the first assertion, which is
why the rate moves between runs and between machines.

The same reading is the benchmark memory report's host baseline
(`bench/bench_util.c:475`). In practice that call happens well after startup,
so no recorded sweep should be affected, but it reads 0 the same way.

nclack:
`platform_resident_memory` returned 0 both when the read failed and when
the kernel reported no resident pages. `test_resident_memory_is_plausible`
asserted `resident > 0` to check the read worked, and on Linux 6.8 a young
process reads 0 until enough pages push a batch through the kernel's
per-CPU counters, so the test raced process startup.

Both readings now write through an out-parameter and return 0 on success,
so a caller can tell a failed read from a real zero. The test checks the
status instead of the value. The benchmark carries the same distinction
into its console report and its results file, which gains a
`memory_host_reading_failed` field.

Closes #209.

### DIFF
--- a/bench/bench_report.c
+++ b/bench/bench_report.c
@@ -38,7 +38,7 @@ print_memory_report(const struct bench_memory* mem)
 {
   char a[32], b[32];
   fputc('\n', stderr);
-  if (mem->host_peak_bytes) {
+  if (!mem->host_reading_failed) {
     format_bytes(a, sizeof(a), mem->host_baseline_bytes);
     format_bytes(b, sizeof(b), mem->host_peak_bytes);
     print_report("  Host memory:   %s at rest, %s peak", a, b);
@@ -334,6 +334,8 @@ print_bench_json_pass(const struct stream_metrics* m,
   jw_uint(&jw, mem->host_baseline_bytes);
   jw_key(&jw, "memory_host_peak_bytes");
   jw_uint(&jw, mem->host_peak_bytes);
+  jw_key(&jw, "memory_host_reading_failed");
+  jw_bool(&jw, mem->host_reading_failed);
   jw_key(&jw, "memory_device_used_bytes");
   jw_uint(&jw, mem->device_used_bytes);
   jw_key(&jw, "memory_measured_bytes");
--- a/bench/bench_report.h
+++ b/bench/bench_report.h
@@ -25,9 +25,10 @@ struct bench_memory
   uint64_t estimate_pinned_bytes; // pinned host bytes on GPU, 0 on CPU
   uint64_t host_baseline_bytes;   // resident before the stream was created
   uint64_t host_peak_bytes;       // most resident memory held during the run
-  uint64_t device_used_bytes;     // GPU: free device memory the stream took
-  // Which of the two above the estimate can be held against: device memory on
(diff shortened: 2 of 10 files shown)


nclack:
Verification, on the cluster where #209 was seen (kernel 6.8.12):

- A fresh process reads resident 0 on a `cpu` compute node: 300 of 300 runs,
  reading `/proc/self/statm` as the first statement in `main`. On the login
  node, same kernel, 0 of 300.
- I could not make the old test fail there today. Insert one log line before
  the read and the reading is already non-zero — 200 of 200 on a 64-core node
  — and the unmodified test passed 200 of 200. The window is real but shorter
  on those nodes than on whatever ran the 4-in-20 failures.
- CPU-only build, 32-core node: 35 of 35 ctests pass, and
  `test_platform_memory` passed 50 of 50 repeats.

The 200 ms wait added to the test binary returned immediately in every run
above.

nclack:
Correcting my earlier comment: the wait loop it describes is gone.

Measuring it on a CoreWeave `cpu` node showed the wait did nothing. It only
slept, and sleeping does not flush the kernel's page counters — the loop ran
to its full timeout in 13 of 20 runs. Those runs still passed, which is the
useful part: faulting in the 64 MiB block is what moves the reading, because
16384 pages is far past any per-CPU batch. The wait was costing up to 2 s per
run and buying nothing, so I removed it. The block size is now the thing the
comparison tests rest on, and the constant says so.

Issue #209 reports "touching pages does not bring it forward", measured with
1 MiB. At 256 pages that sits around the batch threshold on a large node,
which fits both results.

After removing it, on a `cpu` compute node: 200 of 200 runs of
`test_platform_memory` pass, repeated across three separate builds, and the
CPU-only suite is 35 of 35.

PR #216, opened 2026-08-21
Correctness at queue depth one

### LINKED ISSUE #212: io: correctness at queue depth one
The filesystem write path is correct only because one worker runs jobs in the
order they were posted. The fences rest on something stronger than FIFO:
`retired_seq` has to be an inclusive watermark, because `shard_delivery.c`
overwrites a shard's fence while an older one may still be unwaited. Raise the
worker count with that assumption in place and a fast footer retiring ahead of a
slow data write publishes a watermark past unfinished work — a caller then
reuses a pinned buffer mid-write and the shard tears, silently.

This step removes the assumption while keeping one worker, so step 3 can raise
the count safely.

Scope and design: `dev/io-scheduler.md`, section "2. Correctness at queue depth
one".

**Done when:** output is byte-identical and failure behavior is unchanged.

Part of #178.

### LINKED ISSUE #217: Destroying an io queue can free a fence waiter's lock
A thread inside `io_event_wait` can still be using the queue’s mutex and
condition variable after `io_queue_destroy` has freed them.

`io_queue_destroy` sets the shutdown flag, wakes the worker, joins it, and then
frees the ring, the mutex and both condition variables. It never counts the
threads waiting in `io_event_wait`, and it never broadcasts `cond_retired`, which
is the one a fence waiter parks on. So a waiter that has not yet re-acquired the
mutex and returned is left touching freed memory.

In practice the window is small and hard to reach. The worker broadcasts
`cond_retired` on every retirement, so by the time it has drained and exited, a
waiter’s condition is already true and it is on its way out. A test that records
a fence on one thread and destroys the queue on another came back clean over
3200 rounds.

This is not new. `main` has the same shape, in both
`src/zarr/io_queue.posix.c` and `src/zarr/io_queue.win32.c` before they were
collapsed, and #216 does not change it.

Closing the window means counting waiters and having destroy block until the
count reaches zero, plus a `cond_retired` broadcast when the shutdown flag is
set so a waiter whose sequence number will never be reached is not stranded.

nclack:
The filesystem sink is still one write at a time. Correctness no longer
depends on that.

A request is described rather than wrapped in a closure: an operation, a
token naming one open of one shard file, a payload, an offset.
Descriptors and their syscalls are behind a backend interface, so
admission, ordering and retirement belong to the queue and the files to
the backend. The two platform copies of the queue are now one.

Retirement is the load-bearing change. A finished sequence number was
published directly, which means "everything up to here is done" only
because one worker finishes work in the order it takes it. The ring is
now fixed and indexed by sequence number, every slot carries its state,
and on retirement a tail cursor is advanced over the slots already
finished. The watermark follows from the structure rather than from the
ordering, so a quick footer write can no longer be read as a slow one
finished.

Around that: room is claimed before a payload is copied, which makes a
ceiling on queued memory possible; a truncate or close is held until the
writes posted ahead of it on its own file have retired, and is passed
over rather than left to hold up another file; a write is refused if its
file is already closing, and refused again at the backend if its token
has been retired; and a backend may report an outcome after execute has
returned.

Threads parked inside the queue are counted at teardown and drained
before their lock is freed, which closes a use-after-free older than
this change.

Output is byte-identical to main on both benchmarks that write, and
failure behavior is unchanged. The queue tests are rewritten against a
fake backend able to finish requests out of order, with cases for the
orderings the old shape could not reach.

Closes #212. Closes #217.


nclack:
## Motivation

Step 1 (#208) measured the queue depth available and found it: 3.4 shard
files with a write waiting on a four-shard scenario against 0.8 on its
single-shard twin. Step 3 raises the worker count to use it, and that is
when the old retirement breaks. A 4 KiB footer retiring ahead of a slow
48 MiB write would publish a watermark past unfinished work, an
`io_event_wait` would return, the caller would reuse its pinned buffer,
and the shard would tear with no error reported.

## Verification

Every stage was gated the same way on a compute node: a CPU-only build,
the full suite, 90 runs of the three timing-sensitive binaries under
`--repeat until-fail:30`, and a byte comparison against `main`.

Nothing in ctest checks the byte comparison. It runs
`bench_stream_smallepoch_4shards` and `bench_stream_256cube_single` with
`-o` and takes a sha256 of every file in both output trees. It pins
`--chunk-bytes`, because the auto-fit otherwise follows the node's free
memory and the trees then differ between nodes for reasons unrelated to
the change. Coverage is 309 MB of output, 224 writes across 80 files,
and both the borrowed and the copying write path. Two independent runs
on `main` give identical manifests, so a difference would mean
something.

Mutation checks, each applied, confirmed failing, and restored:

- the finished sequence number published instead of the tail walked —
  out-of-order retirement fails
- readiness always true — the barrier cases fail
- room released against the completion's byte count instead of the
  request's — the short-answer case fails
- the byte ceiling ignored — the ceiling case fails
- the generation dropped from token resolution — the stale-token case
  fails, and the payload lands in the wrong file

Four review rounds ran over the branch. The first found six defects,
including a use-after-free in teardown and four data races in the test
backend that ThreadSanitizer confirmed. Rounds two and three each found
a defect inside the previous round's fix. Round four found none.

## Two places the plan was wrong

The plan said a truncate or close is ready once its file has one
outstanding request. That deadlocks: `finalize_shards` posts the footer
write, the truncate and the close on one file without waiting, so two
barriers sit in the queue together and neither is ever the only request
left. The worker parks and `pool_fs_destroy` hangs on the join.
Readiness is posted order within a file instead.

The plan also said a fake backend removes the need for the synchronous
failing-truncate hook. It does not, and the reason is a real gap in
`cpu_stream_flush_body` rather than a limitation of the test. Filed as
#218; `dev/io-scheduler.md` is corrected here.

## Out of scope

The dispatch scan is quadratic in the size of the window. It is free
today, because the filesystem backend finishes inside `execute` and the
scan therefore stops on its first step every time. It stops being free
once a request is answered later or a second worker starts. The fix
belongs with step 3 and is written down in `dev/io-scheduler.md`.

Fault injection still lives in the shipping backend — #219. The byte
ceiling ships unlimited, mechanism tested and number left to step 3.

Issue #218, opened 2026-08-21
A flush misses errors from the work it queued


nclack:
A flush can report success for shards whose truncate or close failed.

`cpu_stream_flush_body` (`src/cpu/stream.body.c`) checks the sink for an error,
then finalizes the shards, then returns. Finalizing posts a truncate and a close
per shard, and the flush neither waits for them nor looks at the sink again. Any
error those raise is invisible to the flush that caused them, so it returns
zero, and `record_finalized` — which only runs when the flush succeeded —
publishes a fence and an append extent for a shard whose size on disk is wrong.

This is why `shard_pool_fs` still carries a synchronous failing-truncate test
hook. The hook exists so a test can make a truncate fail in a way the flush
actually notices: it raises the error and returns non-zero on the caller thread,
before anything is queued. An asynchronous failure would go unseen, which is the
bug above rather than a limitation of the test.

The fix looks small. After the finalize loop, call the sink flush — for the
filesystem pool that is already record-fence, wait, and report the error — and
fold its result in. Then the truncate hook can go away and the failure can be
injected through the backend like every other one.

It does change what a flush means: it would block until the shards it finalized
are on disk. It already waits on both io fences before finalizing, so the
addition covers only the truncate and close of partial shards, which is arguably
what a flush should have been doing.

Found while reviewing #216. Not changed there — #216 is required to leave
failure behavior exactly as it was.

Blocked on #216, which rewrites `shard_pool_fs.c`; the fix deletes the
failing-truncate test hook from that file. Nothing in steps 3 to 5 of #178
depends on this, and it depends on none of them, but step 3 rewrites the same
file, so this is cheaper taken first. #219 is easier afterwards, because this
removes one of the three fault hooks it has to move.

PR #223, opened 2026-08-23
Move the filesystem fault hooks into tests

### LINKED ISSUE #219: Move the filesystem fault hooks out of the shipping backend
Four test-only entry points are compiled into the shipping library.
`io_backend_fs_inject_failure`, `io_backend_fs_inject_block` and
`io_backend_fs_stop` are declared in `src/zarr/io_backend.fs.h`; the fourth,
`shard_pool_fs_set_error`, is in `src/zarr/shard_pool_fs.h` and has one caller,
`tests/test_destroy_midstream.c`. Their state — two one-shot flags, a gate
pointer and a stopped flag — and a one-millisecond polling loop ship with the
library, and `pool_fs_destroy` calls `io_backend_fs_stop` on every teardown for
no reason other than releasing a test block.

#218 adds a fifth, `io_backend_fs_inject_failing_truncate`. It moves the
truncate failure onto the worker rather than deleting the hook, which is the
point of that fix, so this issue has more to carry out rather than less.

The seam to hold them already exists and is already used: `bench/sink_throttled.c`
builds its own `struct io_backend` and hands it to `io_queue_create`. What stops
a test doing the same through the pool is one line — `shard_pool_fs_create`
hard-wires its own backend.

The pool also calls `io_backend_fs_add_file` directly, so a test cannot swap the
backend outright. A decorator fits instead: a wrapper that passes every request
to the filesystem backend except the ones it is asked to fail or block, which
after #218 means `IO_OP_TRUNCATE` as well as `IO_OP_NOOP`.
`shard_pool_fs_create` grows a variant that takes the wrapper, the plain one
passes none, and the hook state moves next to the fake backend in `tests/`.

The six test files that call `shard_pool_fs_inject_failing_job` and
`shard_pool_fs_inject_blocking_job` keep working, since those helpers stay the
same shape.

Found while reviewing #216, which moved the hooks from `shard_pool_fs.c` into
the backend but did not take them out of shipping code.

Blocked on #216, which rewrites both files this touches. Best taken after #218,
which rewrites them again. Nothing depends on this, so it can wait until after
#178 is finished.

nclack:
Three functions in src/zarr/io_backend.fs.h existed only for tests:
io_backend_fs_inject_failure, io_backend_fs_inject_block and
io_backend_fs_stop. They are gone, along with the two one-shot flags, the
gate pointer, the stopped flag and the polling loop, and pool_fs_destroy
no longer calls stop on every teardown.

In their place, shard_pool_fs_create_wrapped puts a backend of the
caller's own between the pool's queue and its filesystem backend, and
reports the queue it built. shard_pool_fs_create passes an empty wrapper,
so the shipping path is unchanged.

The fault state now lives in tests/test_io_faults.c, which passes every
request through to the filesystem except IO_OP_NOOP, and fails or blocks
on that one. It also carries a filesystem store, since most of these
tests reach the pool through store->create_pool. The six test files keep
the same shape; each now opens its gate in cleanup, because nothing
releases a blocked job at teardown any more.

Closes #219.

### DIFF
--- a/src/zarr/io_backend.fs.c
+++ b/src/zarr/io_backend.fs.c
@@ -24,11 +24,6 @@ struct io_backend_fs
   _Atomic uint64_t files_opened;
   _Atomic uint64_t files_open_now;
   _Atomic uint64_t files_open_peak;
-
-  _Atomic int fail_next_noop;
-  _Atomic int block_next_noop;
-  _Atomic int* block_gate;
-  _Atomic int stopped;
 };
 
 static int
@@ -134,25 +129,6 @@ io_backend_fs_files_open_peak(const struct io_backend_fs* b)
   return atomic_load(&b->files_open_peak);
 }
 
-void
-io_backend_fs_inject_failure(struct io_backend_fs* b)
-{
-  atomic_store(&b->fail_next_noop, 1);
-}
-
-void
-io_backend_fs_inject_block(struct io_backend_fs* b, _Atomic int* gate)
-{
-  b->block_gate = gate;
-  atomic_store(&b->block_next_noop, 1);
-}
-
-void
-io_backend_fs_stop(struct io_backend_fs* b)
-{
-  atomic_store(&b->stopped, 1);
-}
-
 static int
 resolve(struct io_backend_fs* b, struct io_file_token file, platform_fd* fd)
 {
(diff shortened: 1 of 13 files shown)


nclack:
Design notes and verification, kept out of the description.

**Why the seam sits at the pool.** The pool needs the concrete
`io_backend_fs*` for `io_backend_fs_add_file` when it opens a slot, so a
test cannot swap the backend outright — hence a wrapper rather than a
replacement. The test helper also decorates `struct store`, because five
of the six tests reach the pool through `store->create_pool`. A
`store_fs_create_wrapped` would delete that decorator and its five
forwarding functions, but it moves test scaffolding into the shipping
library, which is the direction this change is trying to reverse. Worth a
follow-up issue rather than a change here.

**Left alone on purpose.** `shard_pool_fs_inject_failing_truncate` and
`shard_pool_fs_set_error` are still shipping symbols. Both live a layer
above the backend, #219 named neither, and #218 removes the first.

**Nothing releases a blocked job at teardown now.** That was the point of
dropping `io_backend_fs_stop`, but it makes the gate the test author's
responsibility: a cleanup path that forgets to open it wedges the io
worker instead of failing. Every gated test now opens its gate first
thing in cleanup, and the four that lacked a ctest `TIMEOUT` have one.

**Verification.** CPU-only build 35/35 and L40 GPU build 56/56, both
warning-free. Review also caught two pre-existing cleanup paths in the
flush tests that destroyed the stream before joining the thread still
inside it; those are fixed here.

PR #224, opened 2026-08-23
More write queue depth

### LINKED ISSUE #213: io: more queue depth
One write at a time leaves filesystem bandwidth unused. Measured on xfs over an
md RAID10 of 8 NVMe drives, more queue depth is worth about 1.4x, flat past 16.

This issue raises the worker count and schedules ready files round-robin. It
adds the flags that select that — `--io-writes-in-flight`,
`--io-writes-in-flight-per-file`, `--io-workers`, `--io-backend` — records them
in the results file, and picks the byte ceiling from step 1's
`io_queued_bytes_peak`.

The rest of step 3 is filed separately. #225 indexes the open-file table and has
to land first, because dispatch scans the window under the queue's lock and the
second worker is what makes that expensive. #227 pre-sizes shard files, without
which depth per file above one buys nothing. #226 gives the sweep a depth to
sweep, and #228 publishes what the depth was worth.

Needs #212, merged as `54b03fa`, and #225.

Scope and design: `dev/io-scheduler.md`, section "3. More queue depth".

**Done when:** the reef-l40 XFS matrix passes.

Part of #178.

nclack:
The filesystem sink ran one write at a time, so however deep the backlog only
one request reached the drive. It now runs several.

The io queue takes a worker count, a ceiling on requests in flight across every
shard file, and a second ceiling per file. Files take turns, so a backlog on
one shard file cannot keep the others idle. The dispatch scan is gone: each
file keeps its requests on a list of its own, and readiness is one comparison
against the head of that list plus a flag for a barrier already running. The
open-file table is indexed by the token's file index rather than searched, and
the table io_queue_stats.c kept alongside it is merged into it.

A backend hands a file index back when the close naming it runs, which is
before that close retires, so a new file naming the same index takes the entry
over and the old close retires against a generation that no longer matches.

Shard files are pre-sized with ftruncate when more than one write per file may
run, because a write that extends a file takes the file's lock for itself on
xfs. The size is every chunk at its worst-case compressed size plus the footer,
and finalize already trims it back. Not on Windows, where only writing moves a
file's valid data length.

New flags --io-workers, --io-writes-in-flight, --io-writes-in-flight-per-file
and --io-backend select all of it, and the results file records what the run
resolved them to. New sweep tier iodepth covers 1 to 32 writes in flight, at
one and at four per file. Results schema goes to 7: no timing from a run with
filesystem output is comparable across the bump.

Defaults are eight workers, eight writes in flight and four per file. The
byte credit is 2 GiB, above the deepest backlog measured.

Closes #213. Part of #178.


nclack:
## Correction: the sink is not producer-starved, and this node is not that array

My earlier comment said none of the scenarios comes near the array, and read
the 2.6 GiB/s as the pipeline failing to feed the drive. That was wrong, and it
would have sent the next step looking in the wrong place.

**An L40 node's `/tmp` is `md127 : active raid1 dm-0[1] dm-1[0]`** — a mirror of
two encrypted device-mapper volumes, so a write goes to both members and
striping buys nothing. The plan's 12-17.5 GB/s table is an md RAID10 of eight
NVMe drives, which is different hardware.

`dev/io-depth/io_depth`, the tool that produced that table, run on this node,
48 MiB writes over 8 files, one per file:

| writes in flight | 1 | 2 | 4 | 8 | 16 | 32 |
|---|---:|---:|---:|---:|---:|---:|
| GB/s | 2.24 | 1.49 | 2.27 | 2.59 | 2.43 | 2.11 |

**chucky reaches that ceiling.** `orca2_single` uncompressed writes 2.79 GB/s
at eight in flight against the raw 2.59, with a 1.1 GiB backlog behind it and
`io_wait_ms_mean` at 209 ms. The drive is the limit, not the producer, and the
1.2x is the whole of what the node has to give — the microbenchmark's own gain
over the same range is 1.16x.

**Per-file depth showing nothing is the expected answer.** `orca2_single`
already keeps about twenty shard files with a write waiting, so the depth is
there and taking it per file adds nothing. The microbenchmark can isolate it
where the bench cannot — one file, four workers:

| | GB/s |
|---|---:|
| one write per file | 1.98 |
| four per file, no pre-sizing | 2.10 |
| four per file, pre-sized | 2.18 |

Pre-sizing is worth about 10% on a file written by itself. That is the case
for keeping it on.

## One default worth deciding

Both the microbenchmark and the sweep peak at **eight** writes in flight on
this node and fall off past it; the plan's RAID10 peaks at sixteen. Eight
gets 1.16x here and 1.29x there; sixteen gets 1.08x here and 1.42x there. I
have left the default at sixteen to match the plan, but eight is the better
compromise if L40-class storage is the common case. It is one constant in
`shard_pool_fs.c`.

`dev/io-scheduler.md` and `bench/machines.toml` are corrected in ab4b2f0.

nclack:
## Measurements

Two sweeps are committed: `bench/results/turin-raid10-ab4b2f0-20260824.json` on
an md RAID10 of eight drives (`cpu-turin-gp-l-243-241`, 11.75 GB/s at one write
in flight, 17.5 at sixteen), and `bench/results/reef-l40-130fda8-20260823.json`
on an L40 node, whose `/tmp` is an md RAID1 mirror of two drives. Neither is
encrypted; the difference is drive count and array shape.

`orca2_single` uncompressed, GB/s written, by depth allowed:

| array | 1 | 2 | 4 | 8 | 16 | 32 | best/depth 1 |
|---|---:|---:|---:|---:|---:|---:|---:|
| RAID10, eight drives | 6.12 | 6.89 | 7.84 | 8.22 | 8.08 | 7.90 | **1.34x** |
| mirror, two drives | 2.31 | 2.60 | 2.73 | 2.79 | 2.65 | 2.60 | 1.21x |

Both peak at eight, so the defaults are eight workers, depth allowed eight and
four per file. The deepest backlog measured is 1392 MiB, under the 2 GiB byte
credit.

Compare depths only over a run of several seconds. At the scenario's own 200
frames a run lasts half a second, nearly all of it startup and the closing
flush, and the peak is eight in one pass and sixteen in the next. The frame
count is overridden in the sweep tier for that reason.

Per-file depth is not visible in either sweep: ten to seventeen shard files are
already busy in `orca2_single`, so the depth is there without it, and
`smallepoch_single` runs 0.03 s. On a single file in the microbenchmark it is
1.98 GB/s at one write per file against 2.18 pre-sized at four. Pre-sizing
stays on for that reason.

## Shared file server

`/bio` is an NFSv3 mount with sixteen TCP connections over an NVMe store, and
the fastest write target an L40 node has: `orca2_single` uncompressed writes
5.12 GB/s to it against 2.19 to that node's own mirror.

One node's ceiling is near 8.8 GB/s and the server has none: the same point run
from one, two and four nodes at once totals 7.4, 12.3 and 25.0 GB/s. Throughput
stops rising at thirty-two 8 MiB writes in flight, and past that only latency
grows. Buffered writes are no faster and cost ten times the system time, so
O_DIRECT is right here too.

With one stream the sink reaches about half the node's ceiling and with eight
all of it, so the limit is the producer rather than the mount:

| `orca2_single` streams at once | 1 | 2 | 4 | 8 |
|---|---:|---:|---:|---:|
| total GB/s | 3.86 | 5.41 | 6.78 | **8.78** |
| depth reached, each | 12.6 | 10.3 | 9.5 | 8.3 |

One stream's queue is starved rather than backed up: the wait is 0.4 ms at
depth 12.6, and the mount is fastest at about thirty-two writes at a time. A
higher cap helps up to sixteen and then flattens, because the depth reached
stays near twelve however high the cap goes. Where the producer stops posting
is not measured, so the starved queue is a reading and the reason for it is
not — see #232. The eight-drive array has the same pipeline limit, widened here
by the mount's higher latency.

Issue #230, opened 2026-08-24
Nine tests hand-build the same store, pool, array and stream


nclack:
Nine tests build the same store, pool, array and stream by hand.

`store_fs_create`, two `mkdirs`, `create_pool`, a `zarr_array_config`,
`zarr_array_create_with_pool`, a `tile_stream_configuration`, and a stream
create — followed by a four-call teardown in the reverse order — appear in:

- `tests/test_destroy_midstream.c` (3)
- `tests/test_flush_drain_gpu.c` (3)
- `tests/test_destroy_drain.c`
- `tests/test_flush_drain_cpu.c`
- `tests/test_flush_error_drain_cpu.c`

Only the dimensions, the dtype, and one or two config fields differ between
copies. The teardown order is load-bearing — the stream reads sink buffers, so
it has to go before the array, the pool, and the store — and it is retyped by
hand in all nine.

The cost shows up whenever the construction contract moves. A change to how a
stream is opened, or to what a pool needs at create time, is a nine-place edit
across five files, and a copy that is missed fails as a use-after-free at
teardown rather than as a compile error.

`tests/test_zarr_helpers.h` already owns the shape of the fix with
`test_zarr_sink_open`, but it calls `zarr_array_create()` without an explicit
pool and never hands one back. The tests above all need the pool, because that
is what carries the `shard_pool_fs_inject_*` fault hooks.

Found while reviewing #218, which added the ninth copy.

nclack:
A pool-owning sibling to `test_zarr_sink_open` covers all nine.
`test_zarr_sink_open_with_pool` would take the store path, array name, dims,
dtype and codec, create the pool itself, and hand back `{store, pool, array}`
so a test can still reach the pool for the `shard_pool_fs_inject_*` hooks. A
matching close does the teardown in the one order that is safe.

The stream create stays in each test. Its config is where the cases actually
differ, and the CPU and GPU streams have different create calls.

Best taken after #219, which moves the fault hooks out of the shipping backend
and changes what a test needs the pool for.

PR #233, opened 2026-08-25
Collapse the three one-shot io fault flags into one

### LINKED ISSUE #231: Collapse the one-shot fault flags into one
The fault backend keeps one one-shot flag per pair of operation and
effect. `tests/test_io_faults.h` has `fail_next_noop` and
`block_next_noop`, both applying to `IO_OP_NOOP`. A third,
`fail_next_truncate`, sits in `src/zarr/shard_pool_fs.c` today, and PR
#222 moves it into `struct io_backend_fs`, whose header comment then has
to spell out which op each of the three flags applies to.

One pair of fields does the same job: the op to act on, and whether to
fail that request or hold it at a gate. `faults_execute` already reads
`req->op`, so failing the next `IO_OP_TRUNCATE` costs no more than
failing the next `IO_OP_NOOP`. The flags are separate only because each
hook was added on its own.

The truncate case then needs no shipping code at all. #222 moved that
hook to the backend so it fails on the io worker instead of the caller's
thread. The test wrapper also runs on the io worker, so it can hold the
hook, and `shard_pool_fs_inject_failing_truncate` and the per-slot
pointer behind it can go with it.

Sequencing: this comes after both #219 and #218. #219 (PR #223) takes the
two noop flags out of the shipping backend and puts them in
`tests/test_io_faults.c`. #218 (PR #222) adds the truncate flag to that
same backend. The two move in opposite directions in the same file and
will conflict, so #223 should merge first — then #222 has a test-side
wrapper to put its truncate hook in, and never has to add a flag to the
shipping backend at all. If they merge the other way round, this issue
absorbs the leftover: taking the truncate flag back out of
`io_backend_fs`. Nothing depends on it either way.

nclack:
The test fault backend kept one one-shot flag for every operation and
effect it could inject: fail_next_noop, block_next_noop, and
fail_next_truncate. Failing a fourth operation would have meant a fourth
flag, and the header comment had to say which operation each flag applied
to.

One armed fault replaces all three. It holds the operation to act on and
the effect to apply, packed into one word so a request reads both
together. The injection functions keep their names and signatures, so
every test arms the same fault at the same point.

The wrapper no longer special-cases IO_OP_NOOP. The filesystem backend
behind it already completes a noop, so any request the fault does not
claim goes straight through.

Closes #231

### DIFF
--- a/tests/test_io_faults.c
+++ b/tests/test_io_faults.c
@@ -18,9 +18,11 @@ faults_execute(void* ctx,
 {
   struct io_faults* f = (struct io_faults*)ctx;
 
+  uint16_t armed = atomic_load(&f->armed);
   uint8_t fault = IO_FAULT_NONE;
-  if (req->op == atomic_load(&f->fault_op))
-    fault = atomic_exchange(&f->fault, IO_FAULT_NONE);
+  if ((armed >> 8) == req->op &&
+      atomic_compare_exchange_strong(&f->armed, &armed, (uint16_t)0))
+    fault = (uint8_t)(armed & 0xff);
 
   if (fault == IO_FAULT_FAIL) {
     log_error("io_faults: injected failure of io op %u", (unsigned)req->op);
@@ -171,13 +173,10 @@ Fail:
 
 // --- Injection ---
 
-// The op is stored first so a request already running cannot claim the fault
-// against whichever op was armed before it.
 static void
-arm(struct io_faults* f, uint8_t op, uint8_t fault)
+arm_fault(struct io_faults* f, uint8_t op, uint8_t fault)
 {
-  atomic_store(&f->fault_op, op);
-  atomic_store(&f->fault, fault);
+  atomic_store(&f->armed, (uint16_t)(((uint16_t)op << 8) | fault));
 }
 
 static int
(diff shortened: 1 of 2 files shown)


nclack:
One word rather than two fields. Two fields are read in two loads, so a
request can match the operation from one arming and take the effect from
the next: a worker that has already read `IO_OP_NOOP` will claim the
`IO_FAULT_FAIL` armed a moment later for a truncate, failing the noop and
letting the truncate through. Storing them in a fixed order does not close
that, because the split is on the reading side. Packed into one word, a
request takes both or neither.

The three injection functions stayed as they are. A single
`io_faults_arm(f, op, effect, gate)` would have touched eight test files
for no gain, and #230 is moving those same tests around.

No test arms two faults at once. Six tests arm the blocking noop, one arms
the failing noop, and two arm the failing truncate. The two files that use
more than one kind, `test_store_fs.c` and `test_flush_drain_gpu.c`, use a
separate `struct io_faults` per test function. The hazard above is
therefore latent, not live.

Verified on a 16-core cpu node, RelWithDebInfo with the GPU off: 35 tests
pass before and after, no warnings.

Each fault still fires where it did. The failure log now prints the
operation it fired on: it reads 2, `IO_OP_TRUNCATE`, under
`test_flush_error_drain_cpu`, and 0, `IO_OP_NOOP`, under `test_store_fs`.
Arming each fault with `IO_FAULT_NONE` in turn fails exactly the tests that
depend on it. The truncate fault fails `test_flush_error_drain_cpu`, the
failing noop fails `test_store_fs`, and the blocking noop fails
`test_store_fs`, `test_flush_drain_cpu` and `test_pending_bytes_fs`.

One aside from that last check. `test_pending_bytes_fs` failed with the
block disarmed on one run and passed on another, so its dependence on the
block is racy: with a single worker its writes sometimes pile up fast
enough for the count to be non-zero anyway. That predates this change.

PR #235, opened 2026-08-25
Share the store, pool and array setup in nine tests

### LINKED ISSUE #230: Nine tests hand-build the same store, pool, array and stream

### LINKED ISSUE #237: Five test binaries report success when their setup fails
Five test binaries return success when their setup fails, so ctest reports a
pass for a run that executed no test body.

`main` in each of these files declares `int ecode = 0` and ends with a
`Cleanup:` block that falls through into `Fail: return ecode`:

- `tests/test_destroy_drain.c`
- `tests/test_destroy_midstream.c`
- `tests/test_flush_drain_cpu.c`
- `tests/test_flush_drain_gpu.c`
- `tests/test_flush_error_drain_cpu.c`

Nothing between the two labels sets `ecode`, so every jump out of the setup
returns 0. In `tests/test_destroy_drain.c` the setup has two such jumps:
`CHECK(Fail, test_tmpdir_create(...) == 0)` goes straight to `Fail`, and
`CU(Cleanup, cuInit(0))` goes to `Cleanup` and falls through. A GPU build run
where no device is visible, or any run where the temporary directory cannot be
created, therefore prints a pass having called none of the test functions.

`tests/test_pending_bytes_fs.c` has the shape the others need: `int ecode = 1`
at the top, cleared to 0 only after the setup checks succeed.

nclack:
Nine tests in five files each built a filesystem store by hand, made the
directory the array needs, created a shard pool, filled in an array config,
created the array, and then tore the three down in the one order that is safe.
Only the dimensions, the data type and the codec differed.

test_zarr_helpers gains test_zarr_sink_open_with_pool, which takes a store the
test already made and does the rest. It keeps the pool, because that is what
the fault hooks these tests drive take, and it sizes the pool from the
dimensions the way zarr_array_create does rather than from the fixed count the
nine had copied. The store create stays in the test: six of the nine want the
fault-injecting store from test_io_faults and the other three want the plain
filesystem one.

test_zarr_sink_close now destroys the pool too, so the teardown order is
written once, and all three opens in the helper fail through their own close.
The nine tests reach the sink and its error flag through the accessors the
header already had.

The stream create stays in each test, where the cases actually differ, and the
CPU and GPU streams have different create calls anyway.

Five of these mains declared `int ecode = 0` and ended with a `Cleanup:` block
that fell through into `Fail: return ecode`. Nothing between the labels set
ecode, so any jump out of the setup was reported as success, in a binary that
had run no test body. They now start at 1, cleared once the setup checks pass.

The same family was copied again in test_zarr_s3_sink, with a second copy for
multiscale. The store create is now split out of test_zarr_sink_open and
test_zarr_multiscale_open, so only the lines that build an S3 store are left
there, and the rest comes from the shared openers.

Closes #230
Closes #237


nclack:
Verification, based on `a4b61b4`. The CPU-only build passes 35 of 35 tests
before and after, warning-free, and the GPU configure and build are clean with
56 tests registered in both cases. `clang-format` is clean on the seven touched
files.

I checked the extraction field by field against all nine original setup blocks:
data type, fill value, rank, dimensions, codec, slot base, array name, the
mkdirs, and the array-then-pool-then-store teardown are all unchanged. Two
things do change, and neither is observable. The `mkdirs(".")` the nine did by
hand is gone, because `platform_mkdirp` already creates every ancestor of the
array directory. And the pool now gets as many slots as the geometry needs
instead of a fixed eight; the array only ever indexes slots below that count,
and the slot count feeds nothing but the slot array, so the reuse pattern is
the same.

Left alone on purpose. The new open does not write the root group metadata that
`test_zarr_sink_open` writes, and it does not take a fill value. Adding either
would change what these tests put on disk, and the fault tests are sensitive to
what reaches the io queue. That is also why the two opens stay separate rather
than sharing a body. The stream config stays duplicated in each test, as #230
asked. `io_faults_set_io_scheduling` still works with the helper, since the
store create stays in the test and the setter can go between the two
statements; the hazard #239 describes is untouched here.

Two pre-existing problems turned up while reading these files and are not
addressed. Every one of the five `main()` functions starts `int ecode = 0` and
returns it through the setup failure path, so a test whose temp directory or
CUDA init fails reports a pass having run nothing; `test_pending_bytes_fs.c`
shows the right shape. And `test_zarr_s3_sink.c` carries its own copy of this
whole helper family, which no fix here reaches.

nclack:
Verification for the two follow-up commits.

**The ecode fix.** With `mkdtemp` forced to fail through an `LD_PRELOAD` shim,
so `test_tmpdir_create` returns -1 and the setup check jumps to `Fail`:

| binary | before | after |
| --- | --- | --- |
| test_flush_drain_cpu | exit 0 | exit 1 |
| test_flush_error_drain_cpu | exit 0 | exit 1 |

Without the shim both still exit 0. The "before" numbers come from binaries
built from this branch before the change, kept aside for the comparison.

**The S3 test.** Run against MinIO the way `cpu-tests` does, it passes, and so
does test_store_s3. Binaries built either side of the change were run against
the same MinIO in turn and the two bucket listings, key and size, are
identical. The intermediate-group loop in the shared opener writes nothing
extra here because every array name in this file is a single path component,
and the root group was already written by the copy.

One deliberate difference: the copy set `throughput_gbps = 100` for the array
sink and left the multiscale one at the default 10. Both now take the default.

CPU-only build is warning-free, 36 of 36 pass with `-E "(s3)"`, unchanged from
the branch point.

Issue #232, opened 2026-08-25
One stream reaches half of what a node can write to the shared file server


nclack:
One chucky stream writing to the shared file server reaches 3.9 to 4.8 GB/s,
about half the 8.78 GB/s one node can write to it. Eight of those streams at
once on one node add up to that 8.78, so the shortfall is in one stream rather
than in the mount or the file server.

## Measurement

`/bio`, which is `/mnt/main0`, is an NFS version 3 mount, one mebibyte per
request, sixteen TCP connections (`nconnect=16`), over an NVMe store. Measured
2026-08-25 on `cpu-turin-gp-l-243-179`:

```
bench_stream_orca2_single --backend cpu --codec none --chunk-bytes 256K \
  --frames 2000 --io-workers 32 --io-writes-in-flight 32 \
  --io-writes-in-flight-per-file 4
```

Identical streams of that run were started at once on the one node. Depth
reached is `io_writes_in_flight_mean`, the queue depth a run actually got to,
and `io_run_ms_mean` is how long one write took.

| streams at once | total GB/s | per stream | depth reached, each | io_run_ms_mean |
|---:|---:|---:|---:|---:|
| 1 | 3.86 | 3.86 | 12.59 | 107.9 |
| 2 | 5.41 | 2.71 | 10.25 | 125.4 |
| 4 | 6.78 | 1.69 | 9.48 | 185.1 |
| 8 | 8.78 | 1.10 | 8.27 | 251.3 |

The file server is shared with the rest of the cluster, so a single reading is
an estimate, and the spread across repeats is why the single-stream rate is a
range.

## The node's ceiling

8.78 GB/s belongs to a node rather than to the file server.
`dev/io-depth/io_depth`, writing 8 MiB unbuffered requests from one node,
reaches 8.26 GB/s at 32 writes in flight with a median write of 17 ms, and
8.76 GB/s at 128 with a median of 79 ms. The same point run by one, two and
four nodes at once totals 7.4, 12.3 and 25.0 GB/s. 8.8 GB/s is about 70 Gb/s,
roughly what one of a node's 100 Gb/s ports carries in practice.

## Depth reached against depth allowed

A higher `--io-writes-in-flight` does not close the gap. Depth reached settles
near 13 whatever the cap allows — 13.09 at a cap of 32 and 12.19 at 64 — and
throughput with it, 4.45 and 4.25 GB/s. `io_wait_ms_mean` is 0.4 ms at those
[...]

PR #236, opened 2026-08-25
Report each of the producer's waits on io

nclack:
The CPU path timed one of the producer's waits on io and folded the rest into
the sink stage total, so a starved write queue could be measured but not
explained (#232). Each wait is now reported on its own, the same way
io_fence_stall already was: the wait on a shard's previous footer write before
its buffer is filled again, the wait on shards closed since the append extent
was last published, and the wait on every queued write at flush. The first two
sit in shard_delivery.c, which had no way to reach the metrics, so its four
entry points take a nullable stream_metrics pointer and one helper charges the
wall time of a fence wait to whichever metric the caller names. The GPU path
passes NULL, because its delivery runs on the drain worker rather than the
producer and mixing the two timelines in one entry would make the total
meaningless.

Nothing is subtracted from the sink total, so it keeps its meaning and no older
sweep becomes incomparable. Two of the four waits fall outside every stage
total: the append-extent wait runs in the periodic metadata update, and the
slot fence sits between the compress and aggregate timers, so neither was ever
held by a stage. Adding a metric does not need a schema bump by the rule in
scripts/sweep/README.md, so the results version stays at 7; the new keys are
named in summary.py so the overview page keeps them, and the field table in
that README now describes all four.

A new test drives a CPU stream against a sink that can hold either its fences
or its writes and opens. Counting the calls would not catch a metric that
brackets the wrong interval, so each kind of work is held in turn: with slow
fences every wait metric must show at least one hold in its fastest
measurement, and with slow writes and opens none of them may show even one
while the sink stage shows all of them. Every fence wait the sink is asked for
during streaming has to be charged to exactly one metric.

Measured on the shared file server and on the same node's local array, with
repeats; the numbers are in a comment. They say the producer sits on the fence
between the two aggregate slots, and that the two waits this issue suspected
are nothing. No behavior change beyond the added timing, and the design doc
records the result.


nclack:
Measured on `cpu-turin-gp-l-243-191`, CPU backend, `orca2_single`, `--codec none --chunk-bytes 256K --frames 2000 --io-workers 32 --io-writes-in-flight 32 --io-writes-in-flight-per-file 4`. Call counts are the same in every run: 63 slot fences, 80 footer waits, 4 extent publications, 1 flush.

Six repeats to the file server:

| repeat | wall s | GiB/s | depth | slot fence ms | footer ms | extent ms | flush ms |
|---|---:|---:|---:|---:|---:|---:|---:|
| 1 | 9.12 | 3.86 | 10.6 | 4577 | 0.37 | 567 | 23.2 |
| 2 | 7.50 | 4.68 | 11.1 | 3195 | 0.19 | 473 | 30.7 |
| 3 | 7.34 | 4.79 | 12.8 | 3285 | 0.12 | 345 | 13.9 |
| 4 | 9.00 | 3.91 | 11.5 | 4309 | 0.41 | 531 | 41.8 |
| 5 | 8.28 | 4.25 | 11.5 | 3880 | 0.32 | 443 | 21.7 |
| 6 | 8.14 | 4.32 | 11.5 | 3608 | 0.54 | 454 | 18.3 |

Two of the same run to the node's own eight-drive array, for contrast:

| repeat | wall s | GiB/s | depth | slot fence ms | footer ms | extent ms | flush ms |
|---|---:|---:|---:|---:|---:|---:|---:|
| 1 | 5.88 | 5.98 | 12.7 | 5 | 0.20 | 298 | 36.7 |
| 2 | 5.84 | 6.02 | 12.7 | 15 | 0.51 | 281 | 33.8 |

The slot fence — the existing `io_fence_stall`, the wait on an aggregate slot's
previous writes in `src/cpu/pipeline.c` — is 43 to 50% of a run on the mount
and under a third of a percent on the array. Publishing the append extent is
about 5% on either. The footer buffer is a few tenths of a millisecond over 80
waits, and the flush a few tens over one; neither is worth anything.

So the answer to the issue's missing measurement is that the two waits it
suspected are nothing, and the producer sits on the slot fence. There are two
aggregate slots, so a batch cannot be filled until the writes of the batch two
before it have landed: the producer runs two batches ahead of io and no
further. On the mount, where one write costs about ten times what it costs on a
local drive, that ceiling is what the starved queue is made of. The issue's
reading of 0.01 ms over seven calls came from a different point — the
eight-drive array at 4000 frames — not from this configuration.

The append-extent wait is the one that was invisible: it runs in the periodic
metadata update, outside every stage total, so its 5% appeared nowhere at all.

No fix is attempted here. Three repeats each of the base commit `a4b61b4` and
this branch, alternated in one allocation so both saw the same server:

| | median GiB/s | range | median slot fence ms |
|---|---:|---|---:|
| `a4b61b4` | 4.24 | 3.84 to 4.48 | 3897 |
| this branch | 4.25 | 3.91 to 4.32 | 3880 |

The added clocks cost nothing, and the slot fence was already the dominant wait
before this branch measured the others.

The test proof, since a populated metric can still measure the wrong thing:
`test_stall_metrics_cpu` was rerun against two deliberately broken versions of
the timing helper. Recording a zero instead of the measured interval fails on
`footer_buffer: 6 measurements totalling 0 ms`. Widening the footer bracket to
cover the footer write as well fails on `footer_buffer: 6 waits totalling
120.451 ms, wanted under 20`. The unmodified build passes.

nclack:
The flush wait is keyed `flush_writes_ms` / `flush_writes_count`, not `flush_fence_*` as the table above implies; it was renamed during review so it does not sit one word from the unrelated `flush_stall`. The measured numbers are unchanged.

Two corrections to the field notes, both caught in review and both now right in `scripts/sweep/README.md`. `io_fence` is measured on the GPU path too, in `wait_io_fences`. And it is not part of the `sink` stage total: on the CPU it is timed between the compress and aggregate timers, so `io_fence` and `append_extent` both fall outside every stage, while `footer_buffer` and `flush_writes` break `sink` down.

Issue #214, opened 2026-08-21
io: io_uring backend on Linux


nclack:
An io_uring backend on Linux, behind the same interface as the thread backend,
opt-in, falling back to threads when the kernel or the container will not allow
a ring.

At chucky's write sizes one ring reaches the ceiling on its own, and so do
blocking writes, so this step has to earn its place on evidence rather than on
principle.

Needs #213 first.

Scope and design: `dev/io-scheduler.md`, section "4. io_uring on Linux".

**Done when:** the XFS matrix passes, then an NFS matrix before it becomes the
default.

Part of #178.

PR #238, opened 2026-08-25
Close four gaps in the io queue backend interface

### LINKED ISSUE #229: io: close the gaps in the backend interface
The backend interface #216 built is general enough to hold an io_uring backend,
but four things are missing from it. Settling them in their own pull request
keeps #214 to the ring itself.

- A backend has no way to say "not now", which a full submission queue needs.
  Today it can only block a worker or fail the request.
- The queue never calls into the backend at teardown, so a backend with a thread
  of its own has nowhere to stop it.
- The request handed to `execute` is the worker's own copy and dies when the
  call returns, so a backend that finishes later has to copy what it needs.
- A short write is unowned. `platform_pwrite` loops internally so it never
  reports one; a ring does. Retrying the remainder belongs in the backend
  rather than in the queue.

The fake backend in `tests/` can cover all four without a ring.

Scope and design: `dev/io-scheduler.md`, section "4. io_uring on Linux".

**Done when:** each of the four has a case against the fake backend, and the
filesystem backend behaves as it does now.

Part of #178, before #214.

nclack:
The backend interface is general enough to hold an io_uring backend, but
four things were missing from it. Settling them here keeps the ring
itself to one change.

A backend with no room for a request answers IO_BUSY instead of blocking
a worker or failing the request. The queue puts the request back where
it was, keeping its sequence number and its place on its file, and the
worker moves on to whatever else is ready. One case cannot go back: a
close hands its file index over when it runs, so a new open may already
hold that index, and the refused close is on no list to go back to. That
one is finished as cancelled rather than left waiting to be picked up by
nobody.

A backend can carry a stop, which io_queue_destroy calls once, after the
last request has finished and every worker has stopped, so a backend
running a thread of its own has somewhere to stop it. A backend wrapping
another passes it along, which the test backend that can make io fail now
does.

The request handed to execute is the queue's own copy rather than the
worker's, so it stays put until the outcome is reported and a backend
that finishes later can keep reading it instead of copying what it
needs.

Finishing a short write belongs to the backend that saw one, and
io_write_remaining hands back the part still to be done. platform_pwrite
loops internally so the filesystem backend never sees a short write, but
a ring will.

The filesystem backend is untouched. The fake backend in the tests can
now refuse a request, be stopped, keep the request it was handed and
take a write a few bytes at a time, with a case for each.

Closes #229

### DIFF
--- a/dev/io-scheduler.md
+++ b/dev/io-scheduler.md
@@ -394,14 +394,15 @@ too, so evidence is needed first.
 
 **Close four gaps in the backend interface** (#229), as its own pull request
 before any ring is written. The interface step 2 built is general enough to hold
-one, but a backend has no way to say "not now", which a full submission queue
-needs; the alternatives today are to block a worker or to fail the request. The
-queue never calls into the backend at teardown, so a backend with a thread of
-its own has nowhere to stop it. The request handed to `execute` is the worker's
-own copy and dies when the call returns, so a backend that finishes later has to
-copy what it needs. And a short write is unowned: `platform_pwrite` loops
-internally so it never reports one, a ring does, and retrying the remainder
-belongs in the backend rather than in the queue.
+one, and these four were what it was short of. A backend with no room now
+answers "not now" instead of blocking a worker or failing the request, and the
+queue offers that request again with its place in line kept. A backend can carry
+a stop, called once at teardown, so one running a thread of its own has
+somewhere to release it. The request handed to `execute` is the queue's own copy
+and outlives the call, so a backend that finishes later reads it where it lies
+rather than copying what it needs. And a short write is the backend's to finish:
+`platform_pwrite` loops internally so it never reports one, a ring does, and the
+part still to be done is handed back for retrying.
 
 *Done when:* the XFS matrix is green, then an NFS matrix before it is made the
 default. The file server's own numbers are above, so that matrix has a
--- a/src/zarr/io_backend.h
+++ b/src/zarr/io_backend.h
@@ -26,8 +26,8 @@ struct io_backend
   //
   // A write that comes back short is finished here rather than by the queue,
   // by retrying what io_write_remaining hands back. Reporting fewer bytes
-  // than were asked for says the backend gave up, and raising the error flag
-  // it was given is its own to do.
+  // than were asked for says the backend gave up; the queue does not read
+  // that as an error, so the backend raises the error flag it was given.
   int (*execute)(void* ctx,
                  const struct io_request* req,
                  uint64_t seq,
@@ -35,6 +35,6 @@ struct io_backend
 
   // Called once by io_queue_destroy, after the last request has finished and
   // every worker has stopped. A backend running a thread of its own stops it
-  // here. Optional, and a backend wrapping another must pass it along.
+  // here. This is optional, and a backend wrapping another passes it along.
   void (*stop)(void* ctx);
 };
--- a/src/zarr/io_backend.h
+++ b/src/zarr/io_backend.h
@@ -18,15 +18,12 @@ struct io_backend
 {
   void* ctx;
 
-  // A request taken but not finished is good until its outcome is reported.
-  // The completion is good only for the length of the call. Every request has
-  // to be taken eventually. A write that moved fewer bytes than asked is the
-  // backend's to finish, not the queue's. Neither the count nor the status is
-  // read, so a backend that gives up has to raise the error flag itself.
+  // A request is carried out here. Every request has to be taken eventually,
+  // and one taken but not finished stays good until its outcome is reported.
   int (*execute)(void* ctx,
                  const struct io_request* req,
                  uint64_t seq,
-                 struct io_completion* out);
+                 struct io_completion* out); // good only for this call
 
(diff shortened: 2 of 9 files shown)


nclack:
Why the busy answer is shaped this way. A refused request is put back to
waiting rather than held by the worker, so the worker is free for another
file while the backend is full. It keeps its sequence number, so it also
keeps its place in the per-file order and in the round-robin turn; only
the state flips. The worker then waits for the next request to finish,
because whatever the backend has no room for is work it already holds and
finishing it is what makes room. When nothing is in flight there is no
such finish coming, so the retry falls back to a 100 us timer. Requeuing
does not broadcast: waking every parked worker to try a backend that just
said no is a retry storm, and the requeuing worker always comes back to
the request itself.

Alternatives considered. Looping inside the worker until the backend
takes the request is what the interface already allowed, and it blocks a
worker, which is the gap. Parking unconditionally on the next retirement
deadlocks a backend that refuses an idle queue. Letting the queue retry a
short write means the queue tracking bytes written per request, which is
the backend's own bookkeeping.

Why io_write_remaining returns a request rather than a plain triple: a
backend's write path already takes a request, the result composes with
itself for the next attempt, and clearing owned keeps a retry from
freeing a buffer the queue still owns.

Verification. On a cpu node, a clean RelWithDebInfo build with no
compiler warnings and ctest 35/35 (S3 excluded), unchanged from a4b61b4.
test_io_queue grows from 23 cases to 28 and all 28 pass. src/zarr/
io_backend.fs.c has no diff at all, which is the strongest statement
available that the filesystem backend behaves as it did; the wrapper in
test_io_faults.c only forwards a stop that the filesystem backend does
not set.

Each new case was checked against a mutation of the thing it covers.
Removing the request's stability, so execute gets the worker's stack copy
again, fails request_outlives_execute on the pointer identity check.
Dropping the stop call fails backend_is_stopped_at_teardown. Dropping the
offset advance in io_write_remaining fails
short_write_is_finished_by_the_backend. Dropping the requeue, or either
of its two in-flight decrements, stalls the refusal cases: the check
fails and the drain in io_queue_destroy then waits forever on a request
the queue will not dispatch again, so the regression shows up as a hang
rather than a failed assertion. That is the shape of every stall in this
file already, since io_event_wait has no deadline.

Not covered: a backend that finishes a request from a thread of its own.
The fake reads the request on the test thread after execute returned,
which is the same lifetime question, without a second thread in the
fixture. A ring backend in #214 exercises the real thing.

nclack:
A second review round found a stall I had missed, and turned up three
more findings I decided against. Recording both halves.

The stall. A close hands its file index back to the backend when it runs,
which is before it retires, so a new open can claim that index while the
close is still on its way out. `file_request_added` resets the file entry
for the new generation, and the old close is then on no list at all. My
first version put a refused request back to waiting unconditionally,
which in that window strands it: nothing walks a list that reaches it, so
it is never handed over again, `tail` never passes it, and
`io_queue_destroy` waits on it forever. `requeue_job` now reports that it
could not put the request back, and the worker finishes it as cancelled
with a log line instead. A well-behaved backend cannot reach this, since
releasing the index means it ran the close rather than refusing it, but
the failure mode was an unkillable hang, which is worth a guard.
`refused_request_after_the_index_is_reused` covers it: with the guard
removed, 28 of 29 cases pass and that one hangs.

Three findings I declined, with reasons.

Parking the worker on the next retirement was called a deadlock risk and
replaced with an unconditional 100 us sleep. I kept the park. It only
happens when `in_flight > 0`, and work in flight has already been handed
over, so it never needs the parked worker to make progress: a blocking
backend holds it in another worker that will return and retire, and a
deferred one is completed by whoever completes it. Both paths broadcast
`cond_not_empty`. For a ring whose submission queue frees up in
microseconds, waiting out a 100 us timer instead of waking on the
retirement is the difference that matters.

Broadcasting on requeue was proposed so a parked worker can take the
request back. Liveness does not need it, because the requeuing worker
always comes back to the request itself, and it wakes every parked worker
to call a backend that just said no.

`io_queue_create` stopping the backend on each of its failure paths makes
the rule "the queue always stops the backend" instead of "destroy stops
the backend". A caller holding a null result still owns the backend and
can release it, which is what `shard_pool_fs` already does, so this only
adds branches nothing exercises.

Two findings I agree with but am not acting on here. The achieved-depth
metric counts a request for as long as it was with the backend, including
a spell that ended in a refusal, so `writes_in_flight_peak` will read
high once a backend actually refuses; the definition in `types.io.h` is
"handed to the backend and not yet finished", which is literally what is
measured, and no backend refuses yet, so no sweep number moves. Whoever
reads depth in #214 should know. And clearing `barrier_running` on a
refusal is unobservable: the walk stops at the barrier itself, and the
flag is cleared again when the barrier retires. I kept the line because
leaving a "running" mark set on a request that is not running is a lie in
the state that the next change to `job_is_ready` would read. The same goes
for the stop pass-through in the wrapper, which has no coverage because
the only backend it ever wraps sets no stop: it is the example the ring's
wrapper gets copied from.

Verified on a cpu node: clean RelWithDebInfo build, zero compiler
warnings, `test_io_queue` 29 of 29, ctest 35 of 35 with S3 excluded.

Issue #241, opened 2026-08-26
test_zarr_fs_sink reports success when its CUDA setup fails


nclack:
`tests/test_zarr_fs_sink.c` is a sixth binary with the bug #235 fixed in five
others, and it is the worst of the six.

Its `main` declares `int ecode = 0` at line 2284 and ends with `return ecode` at
line 2440. Between them the CUDA setup jumps out:

```c
CU(Cleanup, cuInit(0));
CU(Cleanup, cuDeviceGet(&dev, 0));
CU(Cleanup, cu_ctx_create(&ctx, 0, dev));
```

`CU` goes to `Cleanup:`, which removes the temp directory and falls through into
`Fail: return ecode`. Nothing between the labels sets `ecode`, so a failed setup
is reported as whatever the cases before it accumulated. The `CHECK(Fail, ...)`
on `test_tmpdir_create` at the top has the same ending.

## Why this one is worse than the five

Each of the five started at 1 because it had nothing before its setup: a jump
out left a binary that had run no test body, so the wrong verdict was at least
a verdict on nothing.

Here twelve cases run first and need no CUDA — `test_for_each_intermediate`, ten
metadata cases, and `test_storage_order_validation`. Passing, they leave `ecode`
at 0. A missing or busy device then returns that 0, and the six cases behind the
setup are skipped without a word: `test_pipeline`,
`test_unbounded_metadata_update`, `test_midstream_metadata_update`,
`test_unbuffered_pipeline`, `test_unbuffered_pipeline_multishard` and
`test_pipeline_storage_order`.

The binary is registered with `add_gpu_test_exe`, so it is built only where a
GPU is expected. Every case skipped this way is one that should have run.

## Fix

The one #235 used: start `ecode` at 1 and clear it once the setup checks have
passed.

`grep -l "int ecode = 0" tests/*.c` returns this file alone on `cc157ce`, so it
is the last of the family.

Follow-up to #237, which is closed correctly: its five are fixed.

Issue #240, opened 2026-08-25
The test flush helpers discard the error they are given


nclack:
`test_zarr_sink_flush` and `test_zarr_multiscale_flush` return void and discard
the error their callee reports, so no test can observe a failed flush.

```c
void
test_zarr_sink_flush(struct test_zarr_sink* z)
{
  zarr_array_flush(z->array);
}
```

`zarr_array_flush` returns `a->pool->flush(a->pool)`, and `src/zarr/shard_pool.h`
documents flush as returning non-zero on error.

Two consequences in tests that exist today. `tests/test_zarr_readback.c` flushes
and then returns success with nothing checked in between, so a flush that failed
is reported as a passing test. In `tests/test_zarr_fs_sink.c` the flush is
followed by a file-exists check and a content check, so the verdict rests on
whatever reached the disk — a shard that was truncated but still parses reads as
a pass.

Issue #239, opened 2026-08-25
Fault scheduling is lost when set before the store is created


nclack:
`io_faults_set_io_scheduling` is silently undone when it is called before the
store or pool is created, and a test that gets the order wrong still passes.

`io_faults_store_create` and `io_faults_pool_create` both open with
`memset(f, 0, sizeof(*f))`, and the setter writes only `f->io`. So this
sequence compiles, runs green, and uses the pool's default scheduling:

```c
struct io_faults faults = { 0 };
io_faults_set_io_scheduling(&faults, (struct io_scheduling){
  .workers = 1, .writes_in_flight = 1, .writes_in_flight_per_file = 1 });
struct store* s = io_faults_store_create(&faults, tmpdir, 1);
```

Nothing reports the loss. A zero field means "take the pool's default", so the
wiped setting cannot be told apart from a deliberate one. The tests that ask
for a single worker do so because a blocking job has to hold up everything
queued behind it; with the default eight workers and eight writes in flight it
does not, and the assertion those tests make can hold for the wrong reason.

The one caller today, `tests/test_store_fs.c`, calls the setter after the
create and is correct. The ordering is not stated in `tests/test_io_faults.h`
and nothing enforces it.

nclack:
Two fixes worth weighing. Taking the scheduling as a parameter on both create functions removes the ordering entirely, at the cost of a second argument on eight call sites that do not care. Keeping the setter and having it refuse once the pool exists is smaller, and turns a silent loss into a failure at the call site.

I introduced this in #224, while resolving the merge with #223 — the store path needed the scheduling and its pool is built later, so a setter was the cheaper option at the time. The ordering hazard was not obvious then and is not written down.

Found by review on #235. Related to #237: both let a test report success while measuring something other than what it claims.

PR #242, opened 2026-08-26
Update the io plan against merged work

nclack:
Five pull requests landed in two days and the write-scheduling plan had not
caught up with any of them. Steps 1 to 3 are marked done, the two gaps step 2
left open are marked fixed with the commits that fixed them and with the order
they were actually taken in, which was the opposite of the one the plan
predicted, and the issue list now records what closed each one.

The step 4 section is the substantial change. Its own gate says a ring is worth
writing only if the same depth is not already reached by threads. The four
interface gaps it wanted closed first are closed, and the evidence the gate
asked for is now in the file: a node caps near 8.8 GB/s writing to the shared
file server whatever the writes in flight and whether or not they are buffered,
eight streams on one node total that same 8.78, one, two and four nodes total
7.4, 12.3 and 25.0 so the cap belongs to a node, and the wait on an aggregate
slot's previous writes is 43 to 50 percent of a run against the file server
against under a third of a percent against a local array. There are two slots,
so the producer runs two batches ahead of io and no further. The section now
says the limit is upstream of the backend, and what would change that: a target
on which a node is not the limit, or a producer that can run further ahead. The
step is not deleted.

Three numbers were wrong and are corrected against the committed results files.
The 209 ms wait quoted for eight writes in flight on the L40 is the reading at
one, and at eight it is 139 ms. The 0.4 ms wait quoted for thirty-two on the
eight-drive array is the four-per-file arm while the 59 ms it is compared with
is the one-per-file arm, so both arms are given. The throughput tables did not
say which per-file arm they were drawn from, and now do.

Two provenance notes are added. The four per-wait readings are in the pull
request that measured them and in no committed sweep, because every committed
sweep predates the counters. The second committed pass of the depth tier on the
eight-drive array was never referenced and does not agree with the first: its
peak is at sixteen rather than eight and its gain is 1.32x rather than 1.34x,
which is the plan's own warning about short runs showing up in its own data.

Terms gains a fence, a producer and an aggregate slot, all three used in the
file without being defined, and the entry on per-file depth is softened to
match the measurements below it and to note that it buys nothing on the file
server. Settled decisions records that one default for writes in flight
whatever the target is not settled, and that the microbenchmark path named in
the file is not a path in this repository.

### DIFF
--- a/dev/README.md
+++ b/dev/README.md
@@ -8,7 +8,8 @@ top rather than reading as if it were still pending.
 
 - [gpu-orchestration.md](gpu-orchestration.md) — how the GPU pipeline is put
   together, the rebuild that shipped, and the next three pieces of work.
-- [io-scheduler.md](io-scheduler.md) — why only one block at a time is written
-  by the filesystem sink, what was measured, and the four changes left.
+- [io-scheduler.md](io-scheduler.md) — the filesystem sink's write queue, what
+  was measured, the three steps that shipped, and the evidence against the two
+  left.
 - [devlog.md](devlog.md) — running journal and long-lived TODO list, newest
   entries first.
--- a/dev/io-scheduler.md
+++ b/dev/io-scheduler.md
@@ -1,6 +1,9 @@
 # Filesystem write scheduling
 
-**Status.** Steps 1 to 3 are done. Steps 4 and 5 are planned. Issue #178.
+**Status.** Steps 1 to 3 are done, and so is the groundwork for step 4. The
+evidence step 4 waited on is in, and it is against writing an io_uring backend:
+the limit is upstream of the write path. Step 5 is blocked on hardware.
+Issue #178.
 
 Only one write at a time was run by the filesystem sink: each queued job was a
 blocking `pwrite`, so however deep the backlog, only one request reached the
@@ -12,10 +15,12 @@ drive. This file is the working plan for changing that.
   seen.
 - **Queue depth** — writes in flight across every shard file at once. Always
   both words: bare "depth" already means pipeline depth in `src/gpu/`.
-- **Queue depth per file** — the part of that depth on one shard file. No
-  gain above one while a file is still growing: on xfs the file's lock is taken
-  for itself by an extending write, so requests are accepted by the kernel and
-  then run one at a time. Only worth raising once the file is pre-sized.
+- **Queue depth per file** — the part of that depth on one shard file. There
+  is little gain above one while a file is still growing: on xfs the file's
+  lock is taken for itself by an extending write, so requests are accepted by
+  the kernel and then run one at a time. It is worth raising once the file is
+  pre-sized, and worth nothing on the shared file server, where one file is
+  serialized whatever the depth.
 - **Depth allowed** — the cap `--io-writes-in-flight` sets on queue depth, and
   `--io-writes-in-flight-per-file` on queue depth per file. The two are a
   minimum, not a product: four per file across sixteen files still runs only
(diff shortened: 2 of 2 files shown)


nclack:
## Recommended next sweep

Ordered. Sizes are rough. Everything below was checked against `cc157ce`.

**1. Merge #234, which closes #211.** Written, green on all eight checks, waiting.
It is the only correctness bug left in shipping code: two plate fields of view
alias pool slots, so the second field's open closes the first field's file and
takes its place, and the first field's next write lands in the second field's
file at the first field's offset. Everything else on this list is test quality
or performance. The PR adds a case that opens a shard in two fields of one well
at once and reads both files back, so the bug is reproduced rather than argued.
Size: zero, it is done.

**2. #241, the sixth binary that reports success when its setup fails.** Filed
today as a follow-up to #237. `grep -l "int ecode = 0" tests/*.c` returns
`tests/test_zarr_fs_sink.c` alone, and it is the worst of the six: twelve
CPU-only cases run before the CUDA setup and leave `ecode` at 0, so a GPU
runner that has lost its device returns that 0 and six pipeline cases are
skipped in a run reported as green. Every other item here assumes tests ran;
this one decides whether they did. Size: two lines.

**3. #240, the test flush helpers discard the error they are given.** Still
true: `test_zarr_sink_flush` and `test_zarr_multiscale_flush` are `void` in
`tests/test_zarr_helpers.h`, and there are eleven call sites across five files,
every one of them dropping the result. `tests/test_zarr_readback.c` flushes and
returns success with nothing checked in between. Same class as #241 one level
up, so take them together. Size: small, two signatures and eleven call sites.

**4. #239, fault scheduling wiped when it is set before the store is created.**
Still true: `io_faults_store_create` and `io_faults_pool_create` both open with
`memset(f, 0, sizeof(*f))` and the setter writes only `f->io`. Ranked below #240
because there is one caller today and it has the order right. It earns its place
anyway: a zero field means take the default, so a wiped setting cannot be told
from a deliberate one, and the tests that ask for one worker do so because a
blocking job has to hold up everything queued behind it. With the default eight
workers their assertion holds for the wrong reason. The io tests are going to
multiply. Size: small.

**5. #232, one stream reaching half of what a node can write to the file
server.** Re-scope it before picking it up. Its "Missing measurement" section is
done: #236 timed each of the producer's waits, and the answer is that the wait
on an aggregate slot's previous writes is 43 to 50 percent of a run against the
file server. The comment in `src/cpu/stream.c` it also complains about is
corrected. What is left is the fix, and this PR names the two candidates: more
than two aggregate slots, or a fence per shard file rather than per slot. This
is the only remaining performance work with a measured target, and #214 is
behind it. Size: medium.

**6. Update the #178 checklist.** Its body still shows step 3 open with #226 and
#228 unticked and #229 unstarted. All four are closed. Anyone reading the epic
to find work gets the wrong picture. Size: minutes, no code.

## Not next

**#214, the io_uring backend.** Its premise has changed, which is the substance
of this PR. The gate it set for itself is that a ring is worth writing only if
threads do not already reach the same depth, and eight streams on one node reach
the node's own 8.78 GB/s cap on the thread backend. One stream's missing time is
the producer's fence, not the cost of submitting a write. A ring built now could
not be told apart from the threads. Reopen it after #232 moves the producer, or
close it against the evidence. Size if taken: large.

**#215, overlapped writes on Windows.** Blocked on hardware, and the same
argument holds there with less evidence, since there is no Windows GPU CI and no
Windows measurement at all.

**#168, blosc on GPU.** Accurate — `src/gpu/compress.cu` switches on none, lz4
and zstd only, while `src/cpu/compress_blosc.c` carries the two blosc codecs —
but the issue body is empty and sizing it means first working out whether nvcomp
can produce blosc-framed output. The unknown is the reason to skip it, not the
size.

**#34 and #35, the readback tests.** One line each, from April and June, with no
acceptance criteria and no CI dependency in the tree for either tensorstore or
an S3 reader. The interesting half of #34, shard consistency while metadata
changes mid-stream, is a real question and the tree has better machinery for it
than it did. Neither is specified well enough to hand to anyone yet.

PR #244, opened 2026-08-26
Report the flush error in test helpers

### LINKED ISSUE #240: The test flush helpers discard the error they are given

nclack:
The test helpers test_zarr_sink_flush and test_zarr_multiscale_flush returned
void, so the error their callees report never reached a test.
test_zarr_sink_flush now returns it, and all eleven call sites in
test_cpu_zarr_fs.c, test_zarr_fs_sink.c, test_zarr_readback.c and
test_zarr_s3_sink.c check it. Nine sit between a writer_flush that had to
succeed and a check of what reached the store, so none of them wanted the
error dropped. At the other two the flush return is the only signal:
test_zarr_readback flushed and then returned success with nothing checked in
between, and test_zarr_s3_sink drives the shard sink directly with no writer.
That S3 site also checks more than a drain, because pool_s3_flush finalizes an
upload left open and returns that error.

test_zarr_multiscale_flush is deleted rather than widened. It has had no caller
since it was added, so its return value could never be observed.

A new case in test_flush_error_drain_cpu.c has the shape of a real call site:
it writes an epoch, flushes the writer, makes one queued operation fail, and
requires the sink flush to report it while the shard file such a site would
check is whole on disk. Nothing else asserted that a zarr array flush surfaces
a pool error. It carries a 60 second bound, like the other tests that wait on
the io queue.

Closes #240

### DIFF
--- a/tests/CMakeLists.txt
+++ b/tests/CMakeLists.txt
@@ -207,12 +207,11 @@ if(CHUCKY_ENABLE_GPU)
     )
 endif()
 add_test_exe(test_flush_error_drain_cpu test_flush_error_drain_cpu.c stream_cpu zarr_array store_fs store_api shard_pool_fs platform writer test_platform chucky_log test_io_faults test_zarr_helpers)
+# A fence that never retires would hang rather than fail; bound it.
+set_tests_properties(test-test_flush_error_drain_cpu PROPERTIES TIMEOUT 60)
 add_test_exe(test_pending_bytes_fs test_pending_bytes_fs.c shard_pool_fs platform writer test_platform chucky_log test_io_faults)
-# A worker left held would hang rather than fail; bound them.
-set_tests_properties(
-    test-test_flush_error_drain_cpu test-test_pending_bytes_fs
-    PROPERTIES TIMEOUT 60
-)
+# A worker left held would hang rather than fail; bound it.
+set_tests_properties(test-test_pending_bytes_fs PROPERTIES TIMEOUT 60)
 add_test_exe(test_shape_after_flush_cpu test_shape_after_flush_cpu.c stream_cpu test_shard_sink writer chucky_log)
 add_test_exe(test_stall_metrics_cpu test_stall_metrics_cpu.c stream_cpu test_shard_sink writer platform chucky_log)
 add_gpu_test_exe(test_shape_after_flush_gpu test_shape_after_flush_gpu.c LINKS CUDA::cuda_driver CUDA::cudart_static stream test_shard_sink writer chucky_log)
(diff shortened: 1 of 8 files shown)


nclack:
Verified on a CoreWeave node from `cc157ce`.

CPU-only (`-DCHUCKY_ENABLE_GPU=OFF`, RelWithDebInfo, `ctest -E s3`): 36 of 36
passed, warning-free. Matches the baseline.

GPU (`-DCHUCKY_ENABLE_GPU=ON -DCMAKE_CUDA_ARCHITECTURES=89` on an L40,
`ctest -E s3`): 57 of 57 passed, warning-free. Matches the baseline. This is
the config that compiles the six `test_zarr_fs_sink.c` call sites.

The new error path is load-bearing. With the io fault armed and the new case's
assertion flipped to `test_zarr_sink_flush(&z) == 0`, so that it reads like the
eleven real call sites:

- helper as it was, dropping the error: `exit=0`, the test passes although the
  flush failed
- helper as it is here, returning the error: `exit=1`, with
  `check failed: test_zarr_sink_flush(&z) == 0`

Same binary, same fault, only the helper's return differs.

nclack:
The review pass strengthened the new case after the numbers above were
measured: it flushes a clean pool first, so a flush that always failed cannot
pass, and it drops the unbuffered flag because the case writes no data. The
counts are unchanged, re-measured on the final commit — 36 of 36 CPU-only,
57 of 57 on an L40.

nclack:
The eleven converted sites are now covered, not just the helper's return.

A throwaway patch drove one of the eleven — test_pipeline in
tests/test_cpu_zarr_fs.c — from a store whose io can be made to fail, and
armed a failing operation just before its flush:

- with the new `CHECK(Fail4, test_zarr_sink_flush(&z) == 0)`: `exit=1`, at
  `test_cpu_zarr_fs.c:186 check failed: test_zarr_sink_flush(&z) == 0`
- with that line reverted to the bare `test_zarr_sink_flush(&z);` this PR
  replaces: `exit=0`, the test passes with the flush having failed

Same binary and same fault; only the call site differs. The committed case in
test_flush_error_drain_cpu.c now has that same shape — writer, failing
operation, then a store check that still succeeds — so the pattern is pinned
without a fault injector in four test files.

Counts on the final commit: 36 of 36 CPU-only, 57 of 57 on an L40, both
warning-free.

PR #245, opened 2026-08-26
Take fault scheduling at create time

### LINKED ISSUE #239: Fault scheduling is lost when set before the store is created

nclack:
The field written by io_faults_set_io_scheduling was cleared by
io_faults_store_create and io_faults_pool_create. A scheduling asked for
before the create was silently dropped and the pool left on its defaults. A
zero field is a request for the default, so the loss was indistinguishable
from a choice.

The scheduling is now an argument to both creates, and the setter is gone.
It is held in the store beside the root and the buffering flag until the
pool is built, so no order is left to get wrong. NULL is passed everywhere
the defaults are fine.

A new test is added in tests/test_store_fs.c. A store is created with one
worker and one request in flight, a job is held at a gate, and the eight
writes behind it must still be pending, with the pool's greatest depth at
one. On the defaults neither is true. The FAIL line missing from
test_shard_pool_io_stats is added, and in both tests the gate is now opened
on every teardown path.

Closes #239

### DIFF
--- a/tests/test_destroy_drain.c
+++ b/tests/test_destroy_drain.c
@@ -70,7 +70,7 @@ test_destroy_waits_for_sink_io(const char* tmpdir)
   CHECK(Cleanup,
         test_zarr_sink_open_with_pool(
           &z,
-          io_faults_store_create(&faults, tmpdir, 1),
+          io_faults_store_create(&faults, tmpdir, 1, NULL),
           "0",
           dims,
           3,
--- a/tests/test_destroy_midstream.c
+++ b/tests/test_destroy_midstream.c
@@ -254,7 +254,7 @@ test_destroy_blocks_on_gated_sink(const char* tmpdir)
   CHECK(Cleanup,
         test_zarr_sink_open_with_pool(
           &z,
-          io_faults_store_create(&faults, tmpdir, 0),
+          io_faults_store_create(&faults, tmpdir, 0, NULL),
           "0",
           dims,
           3,
--- a/tests/test_flush_drain_cpu.c
+++ b/tests/test_flush_drain_cpu.c
@@ -76,7 +76,7 @@ test_flush_waits_for_sink_io(const char* tmpdir)
   CHECK(Cleanup,
         test_zarr_sink_open_with_pool(
           &z,
-          io_faults_store_create(&faults, tmpdir, /*unbuffered=*/1),
+          io_faults_store_create(&faults, tmpdir, /*unbuffered=*/1, NULL),
           "0",
           dims,
           3,
--- a/tests/test_flush_drain_gpu.c
+++ b/tests/test_flush_drain_gpu.c
@@ -74,7 +74,7 @@ test_flush_waits_for_sink_io(const char* tmpdir)
   CHECK(Cleanup,
         test_zarr_sink_open_with_pool(
           &z,
-          io_faults_store_create(&faults, tmpdir, 1),
+          io_faults_store_create(&faults, tmpdir, 1, NULL),
           "0",
           dims,
           3,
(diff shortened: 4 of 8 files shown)


nclack:
An argument was chosen over a setter that refuses. Every caller was read
first: eight of io_faults_store_create, two of io_faults_pool_create, one of
the setter. The scheduling was already an argument to io_faults_pool_create,
and room for the bug was left by the difference between the two creates. The
reported case is not caught by a setter that refuses once the pool exists:
at the early call the pool is still NULL, so there is nothing to refuse.
Three things would be needed to catch it that way: a new store-was-created
marker in struct io_faults, a runtime failure path, and a test for that
path, and the caller would still be left with an order to remember. With the
scheduling held in the faults store, beside the root and the buffering flag
already kept there for the later pool build, the field is out of the struct
cleared by the memset, so the loss is impossible rather than reported. The
cost is a NULL at seven call sites where the defaults are fine, each one
argument inside an existing test_zarr_sink_open_with_pool call.

Verification. CPU-only RelWithDebInfo, 36 of 38 registered, the two s3 tests
excluded, warning-free. GPU RelWithDebInfo on an L40 at sm_89, 57 of 57,
warning-free. Both are unchanged from cc157ce, and several users of the
helper are GPU-only test files. The new test was run 20 times without a
failure, its runtime is under a millisecond, and there is no sleep in it.

The test was seen to fail before it was believed. Three mutations were each
built and run ten times on their own, red 10 out of 10:

- The scheduling handed to io_faults_store_create is dropped, standing in
  for the memset. The failure is in the pending-byte assertion.
- NULL is passed by faults_store_create_pool, so the scheduling is stored
  and then ignored at the build. The pending bytes look right and the
  failure is in the depth assertion, which is why both are there.
- The shape before the fix is restored in full: the field back in struct
  io_faults, the setter back, and the test's call to it before the create,
  which is the sequence reported in the issue. The failure is in the
  pending-byte assertion.

One limit is known and measured rather than assumed. Dropping any single
scheduling field on the way to the pool is invisible: each of the three was
zeroed on its own, and the test stayed green 10 out of 10 each time. The
three fields are three ceilings on one quantity and the smallest is the
binding one, so it cannot be identified from a single fixture. Every request
is finished inside the call by io_backend_fs, so the depth is never above
the worker count either. The struct is copied whole by the fix, so an
all-or-nothing loss is the reachable failure, and that is what is covered.
Three fixtures would be needed to tell the three apart, which is a coverage
expansion rather than this fix.

Four things are left alone on purpose, all pre-existing and worth their own
issues:

- Nothing is reported by store_fs_set_io_scheduling when handed any store
  whose create_pool is not the plain filesystem one, which now includes the
  faults store, so it is a silent no-op in exactly the tests where it would
  be used.
- f->pool is never cleared, so the one-pool guard is left reading a pointer
  the caller has already freed.
- There is no NULL check on root in io_faults_store_create, so strbuf_set is
  reached before the root can be refused by store_fs_create.
- The held scheduling literal is now written out three times across two test
  files.

PR #246, opened 2026-08-26
An io_uring write backend on Linux

nclack:
Add an io_uring write backend behind the existing io_backend interface, on
Linux, opt-in through `--io-backend uring`. The worker threads stay the
default, and a run that asks for a ring where none can be had says so once and
writes on the threads instead. The interface needed no change.

The ring takes only writes. It holds one slot per write in flight, hands each
to the kernel and answers that the write will finish later; a thread of its own
reads the ring and reports each one to the queue. A truncate, a close and the
fence marker go to the filesystem backend behind it, which also owns the
open-file table, so those keep the behaviour they had. When every slot is held
the ring answers that it has no room and the queue offers the write again in
its place. A write the kernel only partly carries out is carried on from where
it stopped. At teardown the queue's stop call wakes the reading thread and
joins it.

An entry the ring has been given cannot be taken back, so from the moment one
is prepared the write is in flight whatever the submitting call reports, and a
completion carries the count of times its slot was taken so that one for a
write already dealt with is dropped rather than laid on the write there now.

CMake finds liburing through pkg-config or a plain search, and where it is
missing a stub stands in, so macOS, Windows and a Linux box without liburing
build and behave as before. The CI images gain liburing, and because a
container is refused a ring under the default seccomp profile the backend gets
a run of its own with that profile off; a skip there fails the step.

Tests write the same shards twice, once on each backend, and compare the files
byte for byte, buffered and unbuffered, over two rounds so a writer slot and
the file index behind it are both reused. The rest cover a ring too small for
what it is offered, a stale file token, a write the kernel turns down, a short
write, and teardown with nothing ever posted.

The ring does not earn the default. Paired runs of `orca2_single` put it 1.2%
ahead of the threads on one node and 0.5% behind on another, and 2.5% ahead on
the shared file server with a spread six times that. A write loop with nothing
else in it goes the other way: eight blocking threads reach 13.9 GB/s where one
ring reaches 12.3 at the same depth. Measurements are in a comment below.

Part of #214.


nclack:
Superseded by the measurements below: a third set of paired runs, on a second node, put the ring behind rather than ahead.

nclack:
## The ring against the threads

Both arms are the same binary; only `--io-backend` differs. Runs alternate
threads, uring, threads, uring so a pair is adjacent in time, and the first run
of each set is thrown away. Node: `cpu-turin-gp-l`, 32 cores. Local target
`/tmp`, xfs on an eight-drive md RAID10. Shared target on NFS.

### chucky, `bench_stream_orca2_single --backend cpu --codec none --chunk-bytes 256K --frames 2000`

| target | pairs | threads GiB/s | uring GiB/s | ring / threads | pairs the ring won | t |
|---|---|---|---|---|---|---|
| local array, node A | 10 | 8.67 (sd 0.31) | 8.78 (sd 0.34) | 1.012 | 8 of 10 | +3.1 |
| local array, node B | 5 | 4.98 (sd 0.07) | 4.96 (sd 0.07) | 0.995 | 2 of 5 | -0.6 |
| file server | 8 | 2.27 (sd 0.68) | 2.33 (sd 0.60) | 1.025 | 4 of 8 | +0.5 |

Node A and node B are two machines from the same pool running the same binary.
On A the array kept up and a write took 20 ms; on B it did not and a write took
44. The ring is ahead on one and behind on the other, so the sign follows the
node rather than the backend. The file server moved between 0.90 and 3.14 GiB/s
across its eight pairs, which is the filer's own state.

Neither backend is what the run waits on, and on the local array they look the
same from the queue's side:

| | threads | uring |
|---|---|---|
| writes in flight, mean of 8 allowed | 5.57 | 5.77 |
| one write, mean ms | 19.8 | 20.3 |
| one write, longest ms | 48.9 | 63.6 |
| time waiting to start, mean ms | 13.1 | 13.1 |

### A write loop with nothing else in it

`dev/io-depth/io_depth`, 8 files, 4 writes per file, 2 MiB each, 32 GiB total,
O_DIRECT, on the same local array. The threads arm spends one thread per
worker; the uring arm submits from one.

| depth | threads GB/s | uring GB/s | depth reached | p50 threads | p50 uring |
|---|---|---|---|---|---|
| 8 | 13.90 | 12.26 | 7.99 both | 1.11 ms | 1.32 ms |
| 32 | 14.57 | 13.24 | 31.5 both | 2.90 ms | 4.88 ms |

Three repeats each, spread under 2%. The ring reaches the depth it is asked for
and still writes 10 to 13% less. On the file server the same comparison was
1.29 / 1.20 / 2.04 GB/s for the threads against 0.95 / 1.27 / 2.03 for the
ring — the target moved more between repeats than the arms differ, so nothing
can be read from it. CPU per GB written was 0.09 s either way at depth 8 and
0.104 threads against 0.110 uring at depth 32, so there is no CPU saving to
trade the throughput for.

### Read

A null result, and on the raw loop a small loss. The ring matches the threads
on the pipeline that matters and loses to them when the write path is the only
thing running, at the same queue depth. Nothing here argues for making it the
default, which is why it is opt-in and why the plan doc now says so.

It is still worth keeping. The write path is not what chucky waits on today,
and #236 measured a single stream losing 43 to 50% of an NFS run to a wait that
sits upstream of the backend; when that is gone the write path may be what is
left, and this is then a backend to measure again rather than to write. The same
code is what a registered-buffer or an `SQPOLL` experiment would start from.

### Where it is exercised

A container is refused a ring under Docker's default seccomp profile, so the
ordinary CPU test run skips this backend. The `cpu-tests` job therefore runs it
again with that profile off, and a skip there fails the step instead of passing
quietly. The coverage job runs the same way, so the ring's lines are measured
rather than counted as dead.

### Known limits

Two paths need a kernel that has stopped answering, and both end in a wait
rather than an error. If `io_uring_submit` keeps refusing an entry that is
already in the ring, the write is neither run nor reportable — the entry cannot
be taken back, and calling it failed would free a payload the kernel may still
read. If reading the ring keeps failing, the writes it holds cannot be reported
for the same reason. Both raise the pool's error flag and log.

### Test counts

| config | result |
|---|---|
| CPU only, `-E "(s3)"` | 37 of 39 registered pass, 0 fail |
| GPU on, L40 sm_89, `-E "(s3)"` | 58 of 58 pass, 0 fail |

Baseline on `cc157ce` was 36 of 38 and 57 of 57; the one added entry is
`test_io_uring`. Warning-free in both configs, and 30 repeats of the ring test
plus 20 each of `test_io_queue`, `test_store_fs` and `test_pending_bytes_fs`
found no flake.

### Not `Closes`

Issue #214 is done when the XFS matrix passes and then an NFS matrix, before the
ring becomes the default. One configuration on each filesystem has been run, not
a matrix, and it found nothing to make the ring the default for. The backend,
its fallback, its tests and its measurement are here; the matrix is not, so the
issue stays open.

nclack:
wrong approach

PR #257, opened 2026-09-02
Update nvCOMP to 5.3

nclack:
Update the CUDA container build from nvCOMP 5.1.0.21 to 5.3.0.16. This keeps the GPU build on the current nvCOMP release after confirming that Zstd compression throughput and output sizes remain stable on an L40.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -7,7 +7,7 @@ ARG CUDA_VERSION=13.2.1
 FROM nvidia/cuda:${CUDA_VERSION}-devel-ubuntu24.04
 # Re-declare ARGs needed in subsequent layers; they are reset by FROM.
 ARG CUDA_MAJOR=13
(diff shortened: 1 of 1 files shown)


nclack:
Verification:

- NVIDIA publishes nvCOMP 5.3.0.16 archives for both CUDA 12 and CUDA 13.
- Chucky built from current upstream main with CMake resolving the nvCOMP 5.3 headers and static library explicitly.
- All 58 non-S3 tests passed on one NVIDIA L40.
- Three interleaved Zstd A/B repetitions found no material regression across mixed, zero, and uniform 8-, 12-, and 16-bit inputs. Median paired changes ranged from +0.08% to -0.11% on the uniform inputs. Mixed input measured -0.70%, within its -1.21% to +0.46% run range. Compression ratios and sampled output sizes were unchanged.

PR #260, opened 2026-09-04
Schedule filesystem opens

### LINKED ISSUE #232: One stream reaches half of what a node can write to the shared file server

nclack:
Queue filesystem opens so file creation, truncation on open, and directory
retries run on I/O workers. Reserve file tokens before posting, keep open
flags on the backend, and let each open request own its path until
completion.

Preserve per-file ordering through open, presizing, writes, final
truncation, and close. Live handles are bounded by writer slots plus the
scheduler's request limit (1024 by default). Failed opens report through
the pool error path while queued work drains, and backend destruction
closes any remaining handles.

Tests cover ordering and overlap, path ownership, handle bounds and reuse,
failure cleanup, and stale requests against an open replacement file.

Closes #232

### DIFF
--- a/src/platform/platform_io.darwin.c
+++ b/src/platform/platform_io.darwin.c
@@ -40,7 +40,9 @@ platform_mkdirp(const char* path)
 platform_fd
 platform_open_write(const char* path, int flags)
 {
-  int oflags = O_WRONLY | O_CREAT | O_TRUNC;
+  int oflags = O_WRONLY;
+  if (!(flags & PLATFORM_OPEN_EXISTING))
+    oflags |= O_CREAT | O_TRUNC;
   int fd = open(path, oflags, 0644);
   if (fd < 0)
     return fd;
--- a/src/platform/platform_io.darwin.c
+++ b/src/platform/platform_io.darwin.c
@@ -40,9 +40,7 @@ platform_mkdirp(const char* path)
 platform_fd
 platform_open_write(const char* path, int flags)
 {
-  int oflags = O_WRONLY;
-  if (!(flags & PLATFORM_OPEN_EXISTING))
-    oflags |= O_CREAT | O_TRUNC;
+  int oflags = O_WRONLY | O_CREAT | O_TRUNC;
   int fd = open(path, oflags, 0644);
   if (fd < 0)
     return fd;
--- a/src/platform/platform_io.h
+++ b/src/platform/platform_io.h
@@ -24,11 +24,12 @@ platform_mkdirp(const char* path);
 // Flags for platform_open_write.
 enum
 {
-  PLATFORM_OPEN_UNBUFFERED = 1
+  PLATFORM_OPEN_UNBUFFERED = 1,
+  PLATFORM_OPEN_EXISTING = 2
 };
 
-// Open a file for writing (create/truncate). Returns PLATFORM_FD_INVALID on
-// error.
+// Open a file for writing. PLATFORM_OPEN_EXISTING preserves its contents;
+// otherwise it is created or truncated. Returns PLATFORM_FD_INVALID on error.
 platform_fd
 platform_open_write(const char* path, int flags);
 
--- a/src/platform/platform_io.h
+++ b/src/platform/platform_io.h
@@ -24,12 +24,11 @@ platform_mkdirp(const char* path);
 // Flags for platform_open_write.
 enum
 {
-  PLATFORM_OPEN_UNBUFFERED = 1,
-  PLATFORM_OPEN_EXISTING = 2
+  PLATFORM_OPEN_UNBUFFERED = 1
 };
 
-// Open a file for writing. PLATFORM_OPEN_EXISTING preserves its contents;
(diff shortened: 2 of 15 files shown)


nclack:
Single-L40 validation at commit 736daff reproduced the exact 608-shard warm-NFS case with three alternating runs per revision.

| Warm NFS median | Main | PR |
|---|---:|---:|
| Throughput | 6.611 GiB/s | 6.822 GiB/s |
| Wall time | 4.839 s | 4.690 s |
| Payload-ready to writes-posted | 2256 ms | 103 ms |
| Compression time | 4248.5 ms | 4248.1 ms |

The write-posting interval fell 95.4% to near the 91-94 ms discard controls. Throughput improved 3.2%; all three PR runs (6.749-6.943 GiB/s) exceeded all three main runs (6.353-6.684 GiB/s). Compression changed by less than 0.01%.

Configuration: one NVIDIA L40, orca2_single, GPU Zstd/XOR u16, 16 KiB chunks, 1820 frames, 608 warm NFS shards, ABBAAB order. The file count and representative inode were unchanged before and after, confirming that every timed run reused existing files.

nclack:
Review follow-up:

- The original head bounded simultaneous open calls by worker count, but kept every descriptor live until its shard closed. That was not a constant bound.
- Commit 6e19b0f caps each filesystem shard pool at 64 handles. At the limit it closes an idle handle before opening another; a later write reopens the existing file without creating or truncating it.
- The tests drive repeated reuse above the limit and verify complete file contents, not just the counter.
- Actual initial-open and reopen failures release their claimed slot and let the remaining queue drain. Stale tokens are refused. A dropped close is reclaimed by backend destruction; the Windows test verifies this by deleting the file afterward.

Linux CPU, macOS, Windows, coverage, CUDA 12, CUDA 13, and both Codecov checks pass at 6e19b0f.

PR #266, opened 2026-09-07
cpu: run small scatters on the caller thread

nclack:
Small CPU appends currently dispatch and join the full compute pool even when the scatter contains only a few hundred bytes. Run scatters smaller than 64 KiB on the caller using the existing scatter loop. Larger scatters and compression retain their parallel execution.

Add correctness coverage for a 64 KiB scatter with reordered storage and a 128 KiB input split between inline and parallel calls at a nonzero offset.

The prior instrumented PR264 filesystem experiment on Auk used 512-byte appends, Blosc LZ4/bitshuffle, and three repetitions. With arrivals spaced 100 µs apart, median ordinary-append p99 fell from 394 to 6.0 µs for `smallepoch_single` and from 435 to 7.0 µs for `smallepoch_4shards`. These battery-powered desktop measurements exclude batch boundaries from the ordinary-call statistic; batch processing still blocks the producer. This PR implements the measured dispatch policy without changing pool notification or spinning.

Validation:

- CPU-only Release build.
- All 8 focused CTests pass: transpose, compression, LOD, CPU stream, multiarray, filesystem integration, Python Zarr readback (including Blosc LZ4/Zstd), and flush/drain.
- `clang-format --dry-run --Werror` and `git diff --check` pass.

Targets `main`, which includes merged #264. Benchmark tooling, raw results, and analysis notes remain outside the repository.

### DIFF
--- a/README.md
+++ b/README.md
@@ -184,10 +184,11 @@ multiscale, and multiscale-with-dim0-downsampling modes.
 | `--blosc-block-bytes` | e.g. `16K`, `64K`, `4097` | Required for Blosc | Internal Blosc block size in bytes |
 | `--blosc-shuffle` | `none`, `byte`, `bit` | `none` | Blosc filter; recorded with the level in benchmark JSON |
 | `--reduce` | `mean`, `min`, `max`, `median`, `max_sup`, `min_sup` | `mean` | LOD reduction method |
+| `--full-memcpy-timing` | flag | off | GPU profiling: time every host copy instead of sampling small copies |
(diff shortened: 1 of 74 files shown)


nclack:
This uses an arbitrary size cutoff to size when the scatter should go to a threadpool. Need to assess whether the cutoff is right.

A better short-circuit path would be to detect when the transpose is trivial. Might be able to even save what would turn into a memcpy too if we can just move the pointer.

PR #267, opened 2026-09-07
fs: queue metadata publication behind shard writes

### LINKED ISSUE #268: Separate stream constructor rollback from automatic finalization
Failed stream construction currently calls the same destructor used to finalize a live stream. As a result, `flushed` and `closed` also mean "do not publish anything while unwinding incomplete construction." PR #267 adds multiarray flag sweeps to preserve that distinction when initial metadata publication fails.

Separate resource release from automatic finalization for CPU/GPU single-stream and multiarray implementations. Constructor failure should release acquired resources directly; public destruction should retain its existing automatic flush/close behavior and then release resources. This should remove the construction-only meaning of the runtime flags without adding another persistent lifecycle flag or mode enum.

Concrete examples at the reviewed PR commit:

- [CPU multiarray construction and failure cleanup](https://github.com/acquire-project/chucky/blob/c547000d055ed114ad8bedf13f2b36d884f6ce2c/src/multiarray/stream.c#L419-L477) marks all descriptors closed before initialization and reopens them only after every initial publication succeeds.
- [CPU multiarray destruction](https://github.com/acquire-project/chucky/blob/c547000d055ed114ad8bedf13f2b36d884f6ce2c/src/multiarray/stream.c#L481-L542) mixes automatic finalization with a largely contiguous resource-release tail.
- [GPU multiarray destruction](https://github.com/acquire-project/chucky/blob/c547000d055ed114ad8bedf13f2b36d884f6ce2c/src/multiarray/stream.gpu.c#L341-L376) must preserve CUDA context handling, delivery-worker join, and stream synchronization before freeing storage.

The cleanup also needs to handle allocation failure before the array descriptors exist. Static inspection shows CPU multiarray thread-pool/descriptor allocation failure can reach an unconditional `flush_impl`, which indexes `ms->arrays`. This path also exists at #266's base commit `15eed391`; it is not a new regression from #267, and allocation failure has not been injected to reproduce it. Audit the analogous GPU path as well.

Acceptance criteria:

- Failed construction performs no flush, close, retry, or additional metadata publication. If one array's initial publication succeeds and a later array's fails, cleanup does not republish either array.
- Successful construction still publishes the empty readable extent synchronously before returning, while retaining configured input capacity and bounded append dimensions.
- Live-stream destruction retains automatic finalization, error reporting, and the required order of delivery-worker join, GPU synchronization, sink completion, and resource release.
- Partially initialized and null resources can be released safely, including failure before allocating descriptors.
- Keep `flushed`/`closed` for runtime idempotency. Use private resource-release functions to express constructor rollback.

Validate with the existing CPU/GPU publication initialization-failure cases, focused partial-construction fault injection, and flush/close/destroy coverage for single streams and multiarray streams. Keep this lifecycle refactor separate from #267's ordered-publication simplification.

nclack:
Publishing a closed shard's readable extent waits for filesystem I/O on
the CPU append path or GPU delivery worker. Metadata snapshots now use
the existing scheduler, allowing both paths to continue while publication
waits for earlier writes to succeed.

Metadata replacement is atomic. Flush and close report queued I/O
failures; close waits for publication. New streams advertise an
empty extent while retaining input capacity. Failed constructors release
resources without finalizing sinks.

Small GPU host-copy timings are sampled while preserving copy totals and
append latency metrics. Single-stream benchmark timing includes final
close.

CPU tests, Zarr/NGFF readback, and sweep/report tests pass. CI passes on
CPU, CUDA 12/13, macOS, and Windows.

Closes #268.

### DIFF
--- a/README.md
+++ b/README.md
@@ -184,10 +184,11 @@ multiscale, and multiscale-with-dim0-downsampling modes.
 | `--blosc-block-bytes` | e.g. `16K`, `64K`, `4097` | Required for Blosc | Internal Blosc block size in bytes |
 | `--blosc-shuffle` | `none`, `byte`, `bit` | `none` | Blosc filter; recorded with the level in benchmark JSON |
 | `--reduce` | `mean`, `min`, `max`, `median`, `max_sup`, `min_sup` | `mean` | LOD reduction method |
+| `--full-memcpy-timing` | flag | off | GPU profiling: time every host copy instead of sampling small copies |
 | `-o path` | output directory | omit to discard | Write Zarr output to disk |
 
 Benchmarks report per-stage throughput and latency, compression ratio, memory
-breakdown, and overall pipeline GB/s.
+breakdown, and overall pipeline GiB/s.
 
 ## Architecture
 
--- a/bench/CMakeLists.txt
+++ b/bench/CMakeLists.txt
@@ -85,3 +85,15 @@ foreach(src IN LISTS BENCH_STREAM_SOURCES)
     string(REPLACE ".c" "" target ${src})
     add_bench(${target} ${src} bench_util Zstd::Zstd)
 endforeach()
+
+find_package(Python3 COMPONENTS Interpreter QUIET)
+if(Python3_Interpreter_FOUND)
+    add_bench(test_bench_report test_bench_report.c bench_util Zstd::Zstd)
+    add_test(
+        NAME test-bench_report
+        COMMAND
+            ${Python3_EXECUTABLE}
+            ${CMAKE_CURRENT_SOURCE_DIR}/test_bench_report.py
+            $<TARGET_FILE:test_bench_report>
+    )
+endif()
--- a/bench/bench_gpu.cuda.c
+++ b/bench/bench_gpu.cuda.c
@@ -83,18 +83,19 @@ bench_gpu_report_memory(const struct tile_stream_configuration* config,
   char a[32], b[32];
   format_bytes(a, sizeof(a), mem.device_bytes);
   format_bytes(b, sizeof(b), mem.host_pinned_bytes);
-  print_report("  GPU memory:  %s device, %s pinned", a, b);
+  print_report("  %-17s %s device, %s pinned", "GPU memory:", a, b);
   format_bytes(a, sizeof(a), mem.staging_bytes);
   format_bytes(b, sizeof(b), mem.chunk_pool_bytes);
-  print_report("    staging:   %s   chunk_pool: %s", a, b);
+  print_report("    %-12s %12s   %-12s %12s", "Staging:", a, "Chunk pool:", b);
   format_bytes(a, sizeof(a), mem.compressed_pool_bytes);
   format_bytes(b, sizeof(b), mem.aggregate_bytes);
-  print_report("    comp_pool: %s   aggregate: %s", a, b);
+  print_report(
+    "    %-12s %12s   %-12s %12s", "Compressed:", a, "Aggregate:", b);
   format_bytes(a, sizeof(a), mem.lod_bytes);
   format_bytes(b, sizeof(b), mem.codec_bytes);
-  print_report("    lod:       %s   codec:     %s", a, b);
+  print_report("    %-12s %12s   %-12s %12s", "LOD:", a, "Codec:", b);
   print_report(
-    "    chunks:    %llu/epoch, %llu total (%d LOD levels, batch=%u)",
+    "    Chunks:      %llu/epoch, %llu total (%d LOD levels, batch=%u)",
     (unsigned long long)mem.chunks_per_epoch,
     (unsigned long long)mem.total_chunks,
     mem.nlod,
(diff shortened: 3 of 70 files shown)


nclack:
Supporting implementation and validation details for the squash description.

**Current scope and stack**

Reviewed head: `08f14200524690c52ec37c7d886d4d312712319c`.
Stacked on #266 (`perf/cpu-small-scatter`), base
`9a45f62785e95d4bba282a09e30e77e76e628401`.
Constructor rollback from #268 is included in this head; it is no longer
separate follow-up work. GPU host-copy timing sampling from #270 is also
included.

**Implementation details**

- A queued metadata replacement becomes runnable after every earlier queue
  entry retires. Insertion after backpressure determines that order. Later
  independent file operations can run while publication waits.
- Zarr constructs metadata keys and group envelopes and submits owned
  snapshots. NGFF and HCS supply attributes. Synchronous streaming metadata
  uses the same queue and waits for completion; NGFF submits the child-array
  and group snapshots before waiting. Stores without queued metadata support
  use synchronous writes without flushing active S3 uploads.
- Direct filesystem metadata writes and queued replacements share buffered
  temporary-file writing, parent-directory creation, atomic rename, and
  failure cleanup. Readers see a complete old or new file. There is no added
  `fsync` guarantee; replacements of the same path must be serialized.
- Failed queued I/O suppresses publication and remains a sticky pool error,
  including failures observed by synchronous streaming metadata calls.
  Standalone group writes retain per-call errors. Close and destruction drain
  accepted work; destruction logs failures because it cannot return them.
- New streams publish an empty readable extent while retaining their input
  capacity. Direct array creation retains its configured shape. Constructor
  rollback releases resources without finalizing sinks.
- The benchmark sink wrapper forwards publication and flush hooks. Small GPU
  host-copy timings are sampled, with exact copy totals and separate sampled
  stage reporting; `--full-memcpy-timing` enables full timing. Single-stream
  benchmark timing includes final close. Publication adds no worker thread.

**Validation at `08f1420`**

- [CI run 34651925326](https://github.com/acquire-project/chucky/actions/runs/34651925326)
  passed CPU, CUDA 12/13, macOS, Windows, sweep, and coverage jobs.
- Local macOS review: CPU Debug build and all 41 non-S3 CTests passed,
  including Zarr readback and OME-NGFF validation; all 10 Python sweep tests
  and 4 Node report tests passed.
- The two Python readback validators initially could not download their
  dependencies under restricted network access. Both passed on rerun once
  dependency access was available. GPU and S3 execution were not rerun locally.

**Earlier validation, preserved from the previous description**

These results describe older commits and were not rerun during this review.

At `339b742`, with the then-base `5f0bea9`, validation on Auk reported:

- [CI run 34149879908](https://github.com/acquire-project/chucky/actions/runs/34149879908)
  passed CPU, CUDA 12/13, macOS, Windows, sweep, and coverage jobs.
- All 42 CPU tests passed, including both S3 tests against isolated MinIO;
  all 63 non-S3 tests in the CUDA-enabled build passed.
- Five scheduler, filesystem, metadata, and shared CPU/GPU publication tests
  passed 20 repetitions each. Coverage included ordering under backpressure,
  later independent I/O progress, publication without another append,
  changed/unchanged synchronous extents, sticky failures, snapshot ownership,
  atomic readers, parent creation and replacement cleanup, NGFF/Zarr JSON,
  S3 fallback, teardown, and CPU/GPU multiarray initialization. The shared
  publication test includes the GPU delivery-worker completion assertion.

At benchmark commit `c547000`, 96 paired benchmark executions and API traces
passed. Independent C-Blosc readback verified 84 small-epoch datasets:
5.25 GiB and 5,376 chunks. Final metadata, payloads, and physical shard bytes
matched original base `15eed391` within each backend/layout.

**Historical performance**

These measurements compare original base `15eed391` with feature commit
`c547000`. They predate the later #266 CPU scatter optimization and the
subsequent API, scheduler, backend, timing, and constructor changes.
Benchmarks have not been rerun for the current base or head.

CPU shard-closing append latency, median of three per-run medians:

| Paced case | Base `15eed391` | Feature `c547000` |
|---|---:|---:|
| `smallepoch_single` | 5.18 ms | 1.88 ms |
| `smallepoch_4shards` | 7.12 ms | 1.22 ms |

The runs used 64 MiB of seeded random input, 512-byte appends, 4 MiB batches,
Blosc LZ4/bitshuffle, and enabled metadata publication in both arms. Final
flush and close were timed. Paced GPU append remained about 0.23 ms.
Continuous four-shard throughput rose from 0.69 to 0.92 GiB/s on CPU and
1.80 to 2.41 GiB/s on GPU. Ten continuous single-shard repetitions had
overlapping throughput ranges.

A short `orca2_single` guard rose from 2.88 to 4.29 GiB/s on CPU and
3.04 to 3.40 GiB/s on GPU across three repetitions. It used 64 full frames
(1.125 GiB), with a 288 MiB minimum epoch. These four-batch runs do not
establish sustained production bandwidth.

The previous description recorded raw harnesses, commands, traces, outputs,
and ranges under `/tmp/chucky-async-publication-bench-20260906`, with test logs
under `/tmp/chucky-async-publication-artifacts`. Those local artifacts were
not attached or reverified in this review; this comment preserves the
reported results and caveats, not a publicly accessible raw-data archive.

PR #269, opened 2026-09-07
bench: add microscopy replay benchmarks

nclack:
Stacked on #275; #267 has merged. The measurement foundation in #275 should land first.

`--scenario microscopy` runs all five registered microscopy datasets as six image inputs: OpenCell DNA and protein, BBBC010, JUMP-Scope, DynaCell, and COSEM. Replay preserves native uint8, uint16, and float32 pixels. The manifest loader supports both corpus formats. The submodule points to the public `nclack/chucky-benchmarks-data` repository, with its URL resolved relative to the parent repository. Its six image assets are retrieved by git-annex from GitHub releases; setup no longer requires Reef access.

Image and generated inputs share the measurement lifecycle: preload, fixed geometry, warmup on the measured pipeline, completion of pending work and a metric reset, then measurement through final close with minimum work and timing requirements. Results retain every raw repetition, median throughput, pooled compression ratios, and details from the execution nearest the median. Logical image bytes remain separate from padding. Archived `images` results remain readable; resume checks reject incompatible timing, repetition, and image-content settings. Metadata-only sweep tests use their own fixture and do not require access to the corpus submodule.

The [initial L40 baseline](https://github.com/acquire-project/chucky/blob/bench-microscopy-images/bench/results/reef-l40-432380c-20260912-microscopy-256k.json) adds a 256 KiB sweep across all six inputs, five codec choices, and both backends. All 60 configurations and 180 executions passed, with three executions of at least 32 GiB logical input per configuration. The normal report/explorer retains the results and native pixel types.

Validation for benchmark source `432380c`: focused CPU and GPU correctness checks, independent image readback, sweep/dataset/report checks, and CI on CPU, macOS, Windows, CUDA 12, and CUDA 13 passed. The complete report generated successfully with 35 sweep files and 8,431 runs; its new entry preserves all 60 configurations and 180 raw executions.

The allocation used 58.4 L40 minutes. Median repeat spread was 0.11% on GPU and 1.73% on CPU; seven CPU cases exceeded 5%. Duration/warmup calibration and the chunk-size Pareto study remain follow-up work.


nclack:
Local validation against main passed:

- A clean CPU build without warnings and all 41 CTest cases, excluding S3.
- 65 Python tests covering the sweep runner, image reports, Pareto reports, and corpus tools.
- 17 JavaScript tests covering report selection and Pareto calculations.
- Corpus verification through both command paths, including resolved annex links and independent checksums.
- Image and Pareto browser checks, including input separation, codec labels, repeat tooltips, stage views, mobile layouts, CSV export, and error recovery.

Existing report data and visible browser output matched the main-branch baseline.

nclack:
OpenCell validation is complete on an L40 using the benchmark source at
`e23858a`, built in Release mode for CUDA architecture 89:

- Four targeted CTest entries and 61 independent pixel/input checks passed.
- All 128 executions passed: 32 CPU/GPU cases, each with one warmup and three
  measured replays of at least 8 GiB.
- The report loaded all 32 cases. Maximum throughput repeat spread was 2.46%.
- Core/held-out comparisons met 14 of 16 selection targets. Standalone Zstd
  throughput differed by 10.58% for CPU DNA and 14.46% for GPU protein
  (absolute core/held-out ratio minus one; target at most 10%). All Blosc
  comparisons met the targets. The small corpus remains one confocal source.

Corpus preparation also passed 37 dataset tests, exact pixel checks against
all 16 source planes, fresh-annex retrieval, and materialized ZIP verification.

The latest code change only expands the Pages path filter for
`bench/results/images/**/results.json`. The benchmark implementation is
unchanged from the measured revision. Result files remain uncommitted.

PR #271, opened 2026-09-08
bench: add sustained append throughput and boundary latency measurements

nclack:
## Summary

Add `--duration` and `--warmup` to the existing single-stream, single-scale
discard benchmarks. Fixed-frame runs, pipeline code, and sweep schema remain unchanged.

- Fit geometry from the reference frame count, then stream from a fixed 64 MiB source ring.
- Separate preparation, warmup, measured accepted/completed throughput, and final drain.
- Sample full API calls at batch/generation input boundaries, possible staging boundaries, and the next 16 calls. Label these as input-position groups, not internal event traces.
- Keep whole-run pipeline metrics separate. `--no-boundary-timing` provides an observer-overhead control.
- Retain 80-column text output and add accounting, geometry, CLI, and formatting regression coverage.

No experimental harnesses, result files, or logs are added to the repository.

## Validation

Release builds with and without GPU support. Non-S3 suites: 67 GPU-enabled
tests and 42 CPU-only tests. Boundary timing on/off and fixed-frame compatibility
are covered by the new benchmark integration tests; TTY fixtures cover empty
and very large sample counts.

32 serialized Auk runs at this commit: 5 s warmup, 20 s measured (30 s bulk),
reverse-order repeats, fixed source/layout/budget, none/discard/xor/u16,
performance-core affinity and four workers. Paired mean completed GiB/s:

| Append | smallepoch_single | orca2_single |
|---|---:|---:|
| 512 B | 7.579 | 7.383 |
| 2 KiB | 7.128 | 7.046 |
| 8 KiB | 9.655 | 8.719 |
| 64 MiB | 9.838 | 9.824 |

512 B sampling-on/off mean differences were -0.33% / +0.38%, with pair
differences changing sign on reversal; bulk differences were below 0.003%.

GPU: 3.63 billion recorded append-body calls, maximum 14.93 ms; full-API
boundary maximum 14.85 ms; final-drain maximum 88.04 ms. CPU boundary maximum
59.16 ms. No recorded >=100 ms append events. CPU small-call full-API samples
and GPU wrapper samples do not establish bounds on unsampled calls.

This establishes a sustained baseline, not a production performance change.
The residual small-offer deficit is real in both geometries; the bulk
small-epoch versus Orca gap does not persist.

Stacked on #267, after #270.

### DIFF
--- a/README.md
+++ b/README.md
@@ -184,11 +184,48 @@ multiscale, and multiscale-with-dim0-downsampling modes.
 | `--blosc-block-bytes` | e.g. `16K`, `64K`, `4097` | Required for Blosc | Internal Blosc block size in bytes |
 | `--blosc-shuffle` | `none`, `byte`, `bit` | `none` | Blosc filter; recorded with the level in benchmark JSON |
 | `--reduce` | `mean`, `min`, `max`, `median`, `max_sup`, `min_sup` | `mean` | LOD reduction method |
+| `--duration` | seconds, > 0 | off | Sustained measurement window; single-stream, single-scale discard only |
+| `--warmup` | seconds, >= 0 | 0 | Unmeasured streaming before `--duration` |
+| `--no-boundary-timing` | flag | off | Disable full-API samples to check their observer overhead |
+| `--append-elements` | element count | bulk | In sustained mode, append bytes must divide the fixed 64 MiB source ring |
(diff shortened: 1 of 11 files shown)


nclack:
Superseded by #275, which includes these commits and is now based directly on #267. #269 is adapted and stacked on #275, with one measurement lifecycle for generated and image sources. Merge order: #267 → #275 → #269. Closing only this PR; the branch and its historical measurement evidence are retained.

PR #274, opened 2026-09-09
writer: add optional bounded input buffering

nclack:
Add an optional bounded copying writer so paced producers can reuse accepted
input while CPU/GPU processing continues in a worker.

Drain a shared byte ring in capped batches, releasing capacity after each
batch and continuing immediately while input remains. Preserve input order,
backpressure, and sticky errors through flush and close.

Validated with CPU/GPU readback and sanitizers. The extra copy can reduce
sustained throughput, so buffering remains opt-in.

### DIFF
--- a/README.md
+++ b/README.md
@@ -199,6 +199,8 @@ the library handles all tiling, padding, and shard assembly internally. See
 [docs/design.md][docs-design-md] for a detailed walkthrough, or
 [docs/guide.md][docs-guide-md] for a quick orientation to the module structure.
 
+Paced producers can opt into the [bounded input-buffering writer](docs/buffered-writer.md).
+
 For writing directly to S3 (or S3-compatible stores), see the
 [S3 storage guide][s3-storage-guide].
 
--- a/bench/CMakeLists.txt
+++ b/bench/CMakeLists.txt
@@ -86,6 +86,8 @@ foreach(src IN LISTS BENCH_STREAM_SOURCES)
     add_bench(${target} ${src} bench_util Zstd::Zstd)
 endforeach()
 
+add_bench(bench_buffered_writer bench_buffered_writer.c bench_util)
+
 find_package(Python3 COMPONENTS Interpreter QUIET)
 if(Python3_Interpreter_FOUND AND HAVE_BLOSC)
     add_test(
--- a/bench/bench_buffered_writer.c
+++ b/bench/bench_buffered_writer.c
@@ -18,11 +18,14 @@ struct observed_writer
   int64_t* submitted;
   double* downstream_ms;
   double* completion_ms;
+  double* submission_to_forward_ms;
   size_t frame_bytes;
   size_t frames;
   size_t frame;
   size_t calls;
-  size_t max_batch_frames;
+  size_t call_capacity;
+  size_t bytes;
+  size_t max_batch_bytes;
 };
 
 static struct writer_result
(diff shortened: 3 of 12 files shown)


nclack:
Implementation and validation notes for `e690bc0`.

Paced producers currently pay tile-stream startup and batch flush stalls inside append. This adds an optional bounded copying adapter around `struct writer`. The API is byte-based: it owns accepted input, forwards it in order, and knows nothing about frames or caller append boundaries.

Configuration has two fields:

```c
struct buffered_writer_config config = {
  .capacity_bytes = 128 * 1024 * 1024,
  .max_drain_bytes = 16 * 1024 * 1024,
};
```

`capacity_bytes` bounds all queued and active input. `max_drain_bytes` caps each downstream batch; zero uses the full capacity. Capacity must be nonzero and the cap must not exceed it. Callers choose sizes respecting downstream granularity, such as whole elements.

The producer copies as much input as fits into the shared byte ring, stopping at the ring boundary, and returns the unaccepted suffix. Accepted input can immediately be reused. A drain can split one producer append or combine several: it submits the contiguous queued prefix up to the byte cap in one append, retrying partial downstream acceptance or stalls. It never waits to fill a batch. Each completed batch frees its bytes and the worker immediately continues while input remains. The payload allocation is exactly `capacity_bytes`, plus fixed bookkeeping and one worker thread; there are no per-append descriptors, allocations or extra packing copies.

The adapter borrows downstream and serializes worker access to CPU/GPU tile-stream writers. Producer calls must be externally serialized. Async failures and early completion stop forwarding; unconsumed accepted bytes are counted as abandoned and make finalization fail. Flush drains input and calls downstream flush even after failure. Close publishes metadata after flush. Both are idempotent; destroy joins the worker without destroying the borrowed writer.

Adapter statistics contain byte totals, pending-byte occupancy, backpressure time and batch timing/counts. Per-append slot and queue-timestamp records are removed. The benchmark owns frame-level observation: `submission_to_forward_ms` runs from submission to the first byte's downstream start and includes copying and preceding backpressure; completion ends when the last byte's downstream call returns. These measure consumption, not persistence. `--buffer-bytes` and `--drain-bytes` configure the benchmark, and drains may split benchmark frames. The observer grows its timing array when a small cap creates more batches than frames.

Validation at `e690bc0`: all 64 CPU/GPU CTests passed with temporary MinIO, plus CPU-only adapter/readback checks, AddressSanitizer/UBSan and ThreadSanitizer. Deterministic tests cover a 17-byte capacity with a 5-byte cap, a producer append larger than the cap, exact byte reclamation, ring wrap, source reuse, and failure/early completion inside a split append. CPU/GPU Zarr readback checks byte caps independent of input dimensions and preserves decoded values and metadata. Eight benchmark checks cover CPU/GPU with none/Zstd, a 7 MiB capacity and 3 MiB/768 KiB caps; all completed without abandoned bytes. CLI validation, formatting, diff checks and the documentation example passed.

**Draft: full throughput measurements have not been rerun for this byte-only revision.** At the earlier `e1f24b6` revision, a 128 MiB shared buffer with a cap equivalent to 16 MiB restored GPU Zstd caller stalls near 1 ms at 500 fps and sustained about 1.71 GiB/s. That revision also measured a 10–19% CPU uncompressed throughput regression against `805ed50`. Those are historical results, not current-head performance claims. The copying adapter remains opt-in; a byte cap does not bound downstream call duration or compensate for sustained overload. Current commands and measurement definitions are in `docs/buffered-writer.md`.

PR #275, opened 2026-09-09
bench: unify timing and enforce measurement coverage

### LINKED ISSUE #273: Decouple benchmark geometry from run length and make timing consistent
Benchmark defaults can perform too little work to measure sustained behavior. Increasing frame counts also changes fitted shard geometry, so extending a run changes the workload. Long fixed durations for every case make full sweeps expensive.

Decouple scenario geometry from run length and establish consistent, practical measurement defaults.

Requirements:

- Keep chunk, shard, epoch, and batch geometry fixed as run length changes; support an unbounded frame dimension.
- Exercise and report batch reuse and generation transitions independently of the declared frame count. `min_append_shards` currently serves this timing-coverage purpose.
- Apply the same warmup policy to comparable throughput, stage, and append-latency measurements. Account consistently for in-flight work and final drain; TTY and JSON must agree.
- Validate short-run defaults against longer references without making every sweep case long. Report insufficient coverage rather than implying steady-state accuracy.
- Evaluate whether buffered input improves measurement quality or merely moves warmup into a queue and adds copy overhead; relate this to #272.

Record effective geometry and timing policy for comparison. Defer source-generation changes and generated-data management.

nclack:
Short benchmark runs could pass without exercising buffer reuse, and
their requested length changed fitted geometry. Single-stream benchmarks
now fit geometry separately, warm and drain the pipeline, then reset
observations before measurement.

Accept results only after minimum duration, batch, and generation
coverage. Throughput includes final flush, close, and metadata publication.
Retry excessive final drain with fresh streams, unchanged geometry, and
longer duration; fail when attempts are exhausted.

Reject unqualified sweep results and incompatible resumes. Correct
unbounded multiscale and append-downsample shard metadata. Output retains
warmup data; the two-stream driver keeps its existing policy.

Validated with 44 non-S3 CPU tests, 15 sweep tests, and 4 report tests.

Fixes #273.


nclack:
Reviewed `61e9782`: no actionable findings. The merge preserves constructor
rollback and metric resets after draining.

Validation at this commit:

- macOS CPU Release build and all 44 non-S3 CTests passed.
- All 15 Python sweep tests and 4 Node report tests passed.
- All GitHub checks passed, including CPU, macOS, Windows, CUDA 12, CUDA 13,
  and coverage: [CI run](https://github.com/acquire-project/chucky/actions/runs/34656708625).
- GPU, S3, and the full sweep were not rerun locally.

Warmup requires at least 0.25 seconds and two complete batches. Measurement
requires at least 0.25 seconds, four complete batches, and two generation
transitions. A final drain over 10% of the measured window triggers a longer
attempt using fresh streams and sinks while preserving geometry and source.
The default limit is five attempts. Reports retain the accepted attempt's
rates and observations, its attempt number and duration target, and the time
spent on discarded attempts.

The previously reported full sweep on Auk at `d3edd49` attempted 598
configurations in 117.5 minutes: **547 passed with sufficient coverage;
51 failed**. Failures comprised 45 memory-fit rejections, four GPU allocation
failures, and two `scratch_space_limit` errors. Filesystem cases passed
112/114; local S3 passed 72/72.

The two scratch failures were uncompressed GPU filesystem runs with 32 KiB
chunks (`orca2_single` and `smallepoch_single`). Extended retries hit the
runner's 8 GiB free-space guard before qualifying. Retries can therefore
require substantial temporary storage.

PR #277, opened 2026-09-12
perf: reduce benchmark startup time

nclack:
Limit generated benchmark input to the replay buffer while preserving its
repeat period. Fill CPU reduction tables directly for regular halving and
reuse aligned blocks when constructing scatter and chunk lookup tables.
Measured CPU setup drops from 13.6–84.5 seconds to 0.43–1.05 seconds; total
L40 process time improves by 4.1–9.8x with identical generated input.

Preserve double precision in JSON output so reported durations and their
derived ratios remain consistent. Add deterministic regression coverage
for the drain-fraction CI failure, floating-point round trips, generated
input prefixes, reduction results, and lookup-table geometry.


nclack:
CI validation for `3d1f933` is complete: [all checks passed](https://github.com/acquire-project/chucky/actions/runs/34728961069). Linux passed 52/52 tests; macOS and Windows each passed 50/50; CUDA 12.8.1 and CUDA 13.2.1 each passed 80/80, including the benchmark measurement and filesystem metadata checks. Sweep, coverage, and both Codecov checks are also green.

Startup measurements and implementation details from `b52a4fc` are preserved below. The PR description now also covers the CI corrections in `26fd7dd` and `3d1f933`.

The CUDA 12 failure was caused by the JSON writer formatting doubles with six significant digits. Independently rounding the elapsed duration, drain duration, and their ratio can exceed the measurement test's existing `1e-5` relative tolerance. For example, elapsed `1.170864696` and drain `0.013254351` serialized as `1.17086`, `0.0132544`, and drain fraction `0.0113201`, giving a relative disagreement of `1.11445e-5` when the fraction is recomputed from the JSON durations.

The writer now uses `DBL_DECIMAL_DIG` and retains a decimal point or exponent for floating-point values. The added report fixture reproduces the exact fraction assertion with the old writer and passes with the fix. Separate round-trip checks cover small, large, near-integer, and negative-zero values. The measurement tolerance remains unchanged.

The GPU filesystem metadata test also compared the literal JSON text `0.3`. It now parses the scale array and checks each numeric value, including the default unit scale, so the assertion is independent of decimal spelling.

Generated-data benchmarks prepare 16 complete frames even though replay uses only a 64 MiB buffer. CPU multiscale initialization also rebuilds regular lookup tables with per-pixel coordinate calculations. Bound input preparation to the data actually replayed and reuse the regular structure of those tables.

For medfmt/u16, the temporary generated pattern drops from 4.6875 GiB to 64 MiB while preserving the exact input bytes.

## Implementation

- Generate only the needed XOR/random prefix, retaining the original 16-frame period when it is shorter than the replay buffer. Apply the same bound to the two-stream prefilled buffer.
- For exact halving with the same downsampled dimensions, fill reduction table starts and source indices as arithmetic sequences. Keep the existing builder for odd, clamped, or changing-dimension cases.
- Build CPU scatter and chunk lookup tables in aligned power-of-two blocks of at most 256 elements. Reuse the local offsets within each block; require block sides to divide both image and chunk dimensions. Irregular shapes use the per-element path.

## Measurements

Medians of two repetitions in opposite orders, using four workers. The focused CPU profile measures table construction plus generated-source preparation:

| Scenario | Setup before | Setup after | Speedup |
|---|---:|---:|---:|
| orca2 | 13.607 s | 0.429 s | 31.7x |
| 256cube | 39.598 s | 0.641 s | 61.8x |
| medfmt | 84.502 s | 1.048 s | 80.7x |

Complete L40 benchmark processes, including startup, warmup, measurement, and teardown:

| Scenario | Before | After | Speedup |
|---|---:|---:|---:|
| 256cube_single | 15.372 s | 3.787 s | 4.1x |
| medfmt_multiscale | 33.183 s | 3.381 s | 9.8x |

The paired runs used u16 XOR input, codec `none`, 256 KiB target chunks, 512 MiB target batches, a 16 GiB memory budget, four workers, 0.25 seconds requested warmup, and one second requested measurement. Existing coverage rules extended the measurement windows as needed; all 12 CPU and 8 GPU comparison runs passed coverage and layout checks. The full 64 MiB generated source hashes match between versions for all three profiled scenarios.

CPU process times varied with worker placement. A separate orca2 control restricted to four distinct physical cores reduced median total time from 20.411 s to 8.375 s, while measured processing stayed between 6.7 and 7.1 s across both versions. These short comparisons establish the startup savings; they do not establish a small steady-state throughput change.

## Validation

- All 50 non-S3 CPU tests passed.
- Added independent lookup checks for odd/even shapes, fixed dimensions, storage permutation, padded chunk strides, and serial/parallel execution; brute-force reduction comparisons; and XOR/random prefix checks around frame and repeat boundaries.
- Nine of ten initial L40 checks passed. The remaining measurement check hit its 60-second subprocess timeout in the append-multiscale filesystem case. That focused case subsequently passed against both the unchanged base and this implementation with the original subprocess timeout and all data-readback assertions intact. One diagnostic attempt also reached its separate 100-second overall limit; the final attempt passed in 92.138 seconds with a larger overall allowance for readback and cleanup.
- CPU and CUDA builds completed with GCC 14.3, CUDA 13.1, sm_89, and nvCOMP 5.3.0.16. Formatting and `git diff --check` passed.

PR #288, opened 2026-09-18
Reuse persistent shard footer buffers

nclack:
Write aligned shard footers from their persistent buffers when payloads
use host output groups. The existing completion fence protects reuse,
avoiding an extra allocation and copy for each shard index.


nclack:
Aligned footer buffers already persist between shard generations. Copying each
large shard index into an I/O-owned allocation caused allocator memory to
accumulate during streaming. Borrowing that buffer uses the existing fence
before refilling it, while payload writes retain their output-buffer ownership.

Verification:

- All 51 native CPU tests passed, excluding the two S3 integration tests.
- The new filesystem test passed ten repeated runs. It holds the first footer
  write, verifies that the next generation waits before reusing the buffer,
  then checks both files, indexes, and CRCs.
- The test verifies request ownership for payloads and footers, plus copied
  fallbacks for transient and unaligned footers.
- Removing the ownership fix or skipping the reuse wait makes the regression
  test fail. The restored implementation passes.

PR #289, opened 2026-09-18
Place host buffers near the GPU

nclack:
Prefer memory on the GPU's NUMA node for page-locked buffers, and place
the stream's copy and delivery workers on nearby allowed CPUs. Scope
producer CPU placement to append calls and restore the caller's affinity
and memory policy afterward.

Honor explicit memory policies and current resource limits, with a
portable fallback when NUMA topology or policy support is unavailable.


nclack:
Validation passed in a CPU-only build and on an L40 host with two CPU sockets. The platform, affinity, and threadpool targets passed 3/3. The new GPU placement test and four GPU regression targets passed, covering single/multiarray copying, a restricted cross-thread append and flush, and constructor rollback before and after allocation. All 192 sampled pages across the staging and output buffers were on the selected GPU's NUMA node.

The automatic memory policy prefers local RAM and can fall back under memory pressure. It preserves an explicit caller policy. Append saves affinity in stack storage and performs no placement-related allocation. Independently created filesystem workers and caller-owned input buffers retain their existing placement.

CI can validate the fallback and functional paths on a machine with one NUMA node; it does not need NUMA hardware. No throughput benchmark is included in this change.

PR #290, opened 2026-09-18
Prepare shard files ahead of writes

nclack:
Move shard creation and sizing ahead of delivery when
`config.prepare_shards = 1`. The first generation is ready before stream
creation returns, and one separate worker prepares at most one future
file per slot.

Carry the option through CPU, GPU, multiarray, and NGFF streams. Respect
finite stream limits and remove unused owned files during shutdown,
reporting cleanup failures. Sinks must support both preparation and
cancellation to enable the option.


nclack:
This PR is stacked on #288 (`borrow-shard-footers`) to preserve the tested
footer implementation. Its diff contains the shard-preparation option and
its lifecycle, portability, documentation, and regression coverage.

Preparation uses a separate worker and is excluded from write, metadata,
and footer completion fences. Opening a shard adopts the prepared file;
it waits if preparation has not finished. Flush stops future preparation
before draining delivery, then waits for and removes unused owned files.
Cancellation preserves existing files and directories, reports failures,
and can be retried.

Validation:

- The initial implementation passed 53 CPU tests and all 29 GPU tests.
- After correcting compressed prepared capacity, the final implementation
  passed three focused CPU tests and two GPU tests, including Zarr
  readback and the raw/compressed initial-resize regressions.
- Tests cover empty, partial, complete, finite, and unbounded streams;
  explicit flush/close and destruction; prepared-file adoption; queue
  independence; existing-file preservation; and retryable cleanup.
- Every submitted implementation file matches that final tested version.

